PR #2, opened 2020-11-07
New prevalence analysis examples 

antagomir:
Testing the book build and starting to accumulate material. Requires mia version >=0.0.0.9003 (PR follows in a moment).


FelixErnst:
This will not build correctly on the master branch, since you closed the PR FelixErnst/mia#14 and the `prevalence` function is not available from there.

That the GHA reports a success for this PR, is an error since the bookdown build is actually not run. I fixed this on master, but you have update the GHA action on this branch for it to take effect.

antagomir:
PR on prevalence functions has now been created against FelixErnst/mia. 

It would seem more logical if microbiome/MiaBook is kept in sync with microbiome/mia (and related packages). But this is part of the broader discussion on where the release branches should be located.

microsud:
For the time being, should I commit the additions/changes to `MiaBook` to this branch? So that the discussion can be continued here? Also, set chunks for `prevalence ` as `eval=FALSE` until the `prevalence` PR is merged?

antagomir:
Ok to me.

FelixErnst:
I think would be good to  open as many PR as necessary instead of focussing on a single PR. Just select a topic you want to add and create a branch and subsequent PR for this. Then do a rinse and repeat for any topic, which you encounter. Overlaps will happen, but should not discourage you to add changes in dedicated chunks.

microsud:
Thanks both! The reason I was thinking of an addition to this PR is that I am changing the name of the `data manipulation` to `data operations` and adding info there, plus few lines in `intro`.

FelixErnst:
this can be solved, when merging.

microsud:
Thanks! Will create a new PR then.

FelixErnst:
@antagomir I updated the branch from master and renamed the concepts chapter to microbiome exploration, since for me reading the concepts part is the main aim. Feel free to change it around again

PR #4, opened 2020-12-03
Two chapters initiated.

antagomir:
We can discuss the balance between planning chapters vs. completing chapters. At this stage, I would be willing to add relatively empty chapters so that we can more easily start planning the overall structure.

Some more ideas will follow soon.


antagomir:
Some suggestions on how take the book structure fwd. There is clearly some need to discuss the structure but it may indeed help to add more content first.

13-microbiome-exploration.Rmd
```diff
@@ -31,71 +31,133 @@ Investigating the prevalence of microbes allows you either to focus on changes,
 which pertain to most of the samples, or to focus on less often found microbes,
 which are nonetheless abundantly found in a small number of samples.
 
-Population prevalence (frequency) at a 1% relative abundance threshold:
+On the raw data, the population prevalence (frequency) at a 1% relative
+abundance threshold (`detection = 1/100` and `as_relative = TRUE`), can look
+like this. The low prevalence in this examples can be explained by rather
```
antagomir, line 36: I am not sure what "low prevalence" refers to in here since 1% relative abundance threshold is in fact rather strict for most taxa. I would remove the comment low prevalence, or add some explanation in what sense the prevalence is small. Or we can change the threshold to a less conservative 0.1%

13-microbiome-exploration.Rmd
```diff
@@ -31,71 +31,133 @@ Investigating the prevalence of microbes allows you either to focus on changes,
...
 
 ```{r exploration-prevalence}
 head(getPrevalence(se, detection = 1/100, sort = TRUE, as_relative = TRUE))
 ```
 
-Population prevalence (frequency) at the absolute abundance threshold at read count 1:
+The `detection` and `as_relative` can also be used to access, how many samples
+do pass a threshold for raw counts. Here the population prevalence (frequency) 
+at the absolute abundance threshold (`as_relative = FALSE`) at read count 1
+(`detection = TRUE`) is accessed.
```
antagomir, line 46: This is good and clear text but makes me think where do we draw the line between education book vs. technical manual.. the former emphasizes fluent readability and perhaps shorter expression, the latter emphasizes completeness and exactness. Sometimes it is difficult to achieve both simultaneously. So it is also a general question of the writing style.

13-microbiome-exploration.Rmd
```diff
@@ -31,71 +31,133 @@ Investigating the prevalence of microbes allows you either to focus on changes,
...
+Alternatively the `rank` argument can be set, to perform the agglomeration on
+the fly.
 
+```{r}
+altExp(se,"Phylum") <- agglomerateByRank(se, "Phylum")
+head(getPrevalence(se, rank = "Phylum", detection = 1/100, sort = TRUE,
+                   abund_values = "counts", as_relative = TRUE))
+```
 
+The differnence in the naming scheme, is that by default `na.rm = TRUE` is used
```
antagomir, line 80: differnence -> difference

13-microbiome-exploration.Rmd
```diff
@@ -31,71 +31,133 @@ Investigating the prevalence of microbes allows you either to focus on changes,
...
+for agglomoration in `getPrevalence`, whereas the default for
```
antagomir, line 81: agglomoration -> agglomeration

antagomir (commented):
Seems good.

13-microbiome-exploration.Rmd
FelixErnst: the low prevalence of the samples aka. the result of `getPrevalence`. 0.2 and 0.3 is rather low, wouldn't you agree?

FelixErnst (approved):

13-microbiome-exploration.Rmd
antagomir: "Low" would indicate that the prevalences are lower than we would expect based on a given detection threshold. But in my experience the values are here not specifically low but instead expected when using a conservative detection threshold such as 1% relative abundance.

FelixErnst: I  think in this instance, the situation is different. `GlobalPatterns` contains many different sample types, so that most prevalent ASV has prevalance around `0.3`. I regard this as low, since most of the time if analyzing closely related samples at least a few ASV have a prevalence close to `1.0`. Is my reasoning sound? How would you phrase it differntly so make this point clear?

antagomir: Indeed, heterogeneity in samples makes a difference. Hmm.. but even in that case, isn't this level of prevalence then "as expected" rather than "low"..  It makes more sense now anyway. Low is partially subjective evaluation in this case, one option is to remove but for now, I do not have strong preferences.

FelixErnst: So we leave for now as it is and I merge it?

antagomir: Fine. I may suggest modification at some later revision stage if it still confuses me at that point.

antagomir:
woops - sorry I was too fast to merge - shall I fix the rest?

FelixErnst:
Not for me. I think this is a step in the right direction

PR #7, opened 2020-12-17
Diversity

FelixErnst:
Added diversity calculations, which includes analyses strategies analgous to microbiome package tutorials.

@antagomir Can you add introductions and conclusions to the PERMANOVA and PERMDISP2 as indicated?


antagomir:
ok - will do.

antagomir (approved):
Seems good. We can develop the text further from this.

PR #11, opened 2021-01-20
added text on data import

antagomir:


01-introduction.Rmd
```diff
@@ -72,8 +74,54 @@ In this book, you will come across terms like `FeatureIDs` and `SampleIDs`.
...
 
+
+### Importing data 
+
+
+TODO: data importer from standard CSV/TSV/tree files directly into `TSE`?
+
+Alternative input file formats: data from QIIME2 file formats can be
+directly imported in `TreeSummarizedExperiment` format. See
+`help(mia::makeTreeSummarizedExperimentFromqiime2)` for examples.
```
FelixErnst, line 89: Also phyloseq, biom and DADA2 are supported. My guess would be that they should be mentioned here as well

antagomir:
I was here making a distinction between 1) importing data files to R (QIIME
files) and 2) converting between formats within R (DADA2, phyloseq, biom);
in these latter cases other functions from other packages are needed to
import the data first.

It is potentially confusing that the makeTreeSummarizedExperimentFromqiime2
operates on input files, where as the other functions
makeTreeSummarizedExperimentFrom* operate on R objects. Better to unify
this, and limit the makeTreeSummarizedExperimentFrom* to conversions
between R objects, then have separate commands with a different name
(import* or something) to read data from files directly?

Leo

PR #14, opened 2021-02-22
phylo-div section added

### LINKED ISSUE #12: Add subsection to 6.1 Alpha diversity for Faith’s (1992) index of phylogenetic diversity (PD)
A widely used phylogenetic diversity in microbiome research that can be integrated in this book.

edit: Cannot assign it to myself... as I have limited access to this repo but I can add this sub-section if you are okay with it @antagomir @FelixErnst

microsud:
first draft


14-microbiome-diversity.Rmd
```diff
@@ -51,7 +51,14 @@ by species abundances.
...
   dominance is obtained when one or few species have a high share of
-  the total species abundance in the community.
+  the total species abundance in the community.  
+  
+**Phylogenetic diversity** was first proposed by [@Faith1992], unlike the 
+  diversity measures mentioned above, Phylogenetic diversity (PD) 
+  measure incorporates information from phylogenetic relationships 
+  stored in `phylo` tree between species in a community (sample). The 
+  Faith's PD is calculated as the sum of branch length of all species in 
+  a community (sample).
```
antagomir, line 61: We do not have Faith included in mia::estimateAlphaDiversity at the moment? Should we?

antagomir: Another option would be to make PR to picante, adding `TSE` support.

antagomir (commented):
Looks good except the comment about dependency.

14-microbiome-diversity.Rmd
FelixErnst, line 61: Would be worth pushing the introduction of dominance to a letter point? The order seems a bit confusing 

dominance -> diversity -> alpha/beta/...

Wouldn't this bet better?
diversity -> alpha/beta/... -> dominance

antagomir: Alpha and beta diversity are more different concepts. Dominance is in a way a subset of alpha diversity indices (altough it has an inverse sign). In my opinion, alpha diversity/evenness/richness/dominance/rarity belong together and then beta diversity is an entirely different topic of its own.

antagomir: The order could be: 

richness -> evenness -> diversity -> dominance -> rarity

microsud: Showcasing each type with functions
- Alpha diversity
   - Richness
   - Phylogenetic diversity
   - Evenness
   - Dominance
  
 - Estimating Alpha Diversity
      -  `estimateRichness()`
      -  `estimateDiversity()` # non-phylo
      -  `picante::pd()`           # phylo
      -   `estimateEvenness()`
      -  `estimateDominance()`  Note: can include coverage index in this section
  
 Thoughts?

microsud: If we can improve the speed from `picante` then it would be useful to add. For only one function adding a dependency/suggests is arguable.

antagomir: We could add also "rarity" in the end of this list as it may be added, see [mia#59](https://github.com/FelixErnst/mia/issues/59).

Otherwise ok.

antagomir: Agree. But TSE support in picante itself wouldn't hurt.

14-microbiome-diversity.Rmd
```diff
@@ -97,10 +179,18 @@ plots <- lapply(plots,"+", theme(axis.text.x = element_text(angle=45,hjust=1)))
 ggpubr::ggarrange(plotlist = plots, nrow = 2, ncol = 3, common.legend = TRUE, legend = "right")
 ```
 
+
+Alternatively, all available indices within `mia` can be calculated by default using `estimateAlphaDiversity`...
+
+```{r all-estimateDiv}
+se <- estimateAlphaDiversity(se)
+```
```
FelixErnst, line 187: The use of `estimateAlphaDiversity` and `estimateDiversity` is a bit mixed. Can you align this?

The sentence above should be move up, since the calculation at this point doesn't make that much sense. See line 95ff

microsud: Ah yes. Just to be sure, we expect `estimateAlphaDiversity` to do the following in `mia` package, right?

> All available indices for richness, diversity, evenness, dominance and rarity within `mia` can be calculated by default using `estimateAlphaDiversity`

If yes, then I can introduce it as such.

FelixErnst: Nope `estimateAlphaDiversity` is the same as `estimateDiversity`.

antagomir: I would use `estimateAlphaDiversity` in MiaBook to keep it clear by default. Advanced users can use the shorter form if they want.

antagomir: The `estimateAlphaDiversity` should calculate all available indices. Although dominance indices have an inverse sign, perhaps that does not matter.

FelixErnst: That might be the case, but for this it would need to be clear, how functions are named, what they should perform, how long the run time should be and whether the user can understand a functions purpose.

It does not matter, that we understand it. It must be as clear and concise as it can be. So I have reservations against building a black box, because we want to save two line of code the user hast to type. The two additional lines might be exactly what a user needs to better understand what they are doing and what they can expect from the output. Please keep this in mind.

microsud: > estimateDiversity

at present the man page suggests that `estimateDiversity` is limited to specific indices
```
estimateDiversity(
  x,
  abund_values = "counts",
  index = c("shannon", "simpson", "inv_simpson", "richness", "chao1", "ACE"),
  name = index,
  ...,
  BPPARAM = SerialParam()
)
```

`estimateEvenness` 
```
estimateEvenness(
  x,
  abund_values = "counts",
  index = c("camargo", "pielou", "simpson", "evar", "bulla"),
  name = index,
  ...
)
```
`estimateDominance`
```
estimateDominance(
  x,
  abund_values = "counts",
  index = c("DBP", "DMN", "absolute", "relative", "simpson", "core_abundance", "gini"),
  ntaxa = 1,
  aggregate = TRUE,
  name = index,
  ...,
  BPPARAM = SerialParam()
)
```

While `estimateAlphaDiversity` is more free hand 
```
estimateAlphaDiversity(x, ...)
```
This is why I thought `estimateAlphaDiversity` is parent function to do all calculations.


Therefore, we need to consider how we do this. `estimateDiversity` right now has richness, evenness and diversity indices but not all for these specific concepts. If we do not want `estimateAlphaDiversity` to calculate all indices, then it should not be mentioned and conceptually have individual functions covering specific *type* of indices. This could be more of a choice for our project.

antagomir: Good observation. We have separated the different types of indices and had an idea that there could be a separate wrapper that allows the user to calculate all of these at once. It can be useful because in many studies various richness, evenness and diversity indices are used as complementary views to a general "alpha diversity". In my opinion it would make sense to have such wrapper that binds it all together. But then it must be named so that there is no confusion between pure "diversity" indices (richness+evenness type of thing by definition) and the more generic "alpha diversity" indices that may contain all these variants (even though it is arguable whether dominance belongs to that category due to its different sign).

FelixErnst: I think this warrants a PR in mia to make it clearer and also future proof it, so that additional indices can be added to the correct function.

antagomir: Ok, if we agree that this should be included then we could open an issue and allocate it to someone, then discuss the function naming and possibly other details? This could help to clear up things before going for PR?

14-microbiome-diversity.Rmd
antagomir: `estimateDiversity` is problematic here since it does not specify whether it is alpha or beta diversity. We could use `estimateAlphaDiversity` instead (already implemented)? Although it does match as nicely with other estimate* method names.

14-microbiome-diversity.Rmd
antagomir: I have now opened an [issue #73](https://github.com/FelixErnst/mia/issues/73) to mia

FelixErnst:
Can this be merged or is anyone still working on it?

microsud:
Working on it. Will try to do it once the estimatePD is merged in mia

FelixErnst:
Please also resolve the conflicts, when you work on it. Thx!

FelixErnst (approved):

PR #17, opened 2021-03-05
Example on explained variance to MDS2 and modified beta diversity text

antagomir:
Percent of explained variance is commonly shown for PCA and MDS/PCoA, and I added now an example to MiaBook.

The explained variance is not properly defined for methods such as NMSD, t-SNE or UMAP and cannot be added in the same way unless someone comes up with a proper, unambiguous alternative definition for these (which I am not sure is possible to do due to the mathematical differences).

We can discuss in the PR comments on possible improvements to this workflow.

### DIFF
--- a/14-microbiome-diversity.Rmd
+++ b/14-microbiome-diversity.Rmd
@@ -101,7 +101,17 @@ ggpubr::ggarrange(plotlist = plots, nrow = 2, ncol = 3, common.legend = TRUE, le
 
 Where alpha diversity focuses on community variation within a
 community (sample), beta diversity quantifies (dis-)similarites
-between communities (samples).
+between communities (samples). Some of the most popular beta diversity
+measures in microbiome research include Bray-Curtis index (for
+compositional data), Jaccard index (for presence / absence data,
+ignoring abundance information), Aitchison distance (Euclidean
+distance for clr transformed abundances, aiming to avoid the
+compositionality bias), and the Unifrac distances (that take into
+account the phylogenetic tree information). Only some of the commonly
+used beta diversity measures are actual _distances_; this is a
+mathematically well-defined concept and many ecological beta diversity
+measures, such as Bray-Curtis index, are not proper distances.
+Therefore, the term dissimilarity or beta diversity is commonly used.
 
 Technically, beta diversities are usually represented as `dist`
 objects, which contain triangular data describing the distance between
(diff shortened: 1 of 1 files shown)


14-microbiome-diversity.Rmd
```diff
@@ -132,9 +150,19 @@ Sample similarities can be visualized on a lower-dimensional display
 additional information using variations in colour, shape or size.
 
 ```{r plot-mds-bray-curtis, fig.cap="MDS plot based on the Bray-Curtis distances on the GlobalPattern dataset."}
-plotReducedDim(se, "MDS_BC", colour_by = "SampleType")
+# Create ggplot object
+p <- plotReducedDim(se, "MDS_BC", colour_by = "SampleType")
+
+# Add explained variance for each axis
+e <- attr(reducedDim(se, "MDS_BC"), "eig");
+rel_eig <- e/sum(e[e>0])
```
antagomir, line 158: This calculates explained variance, and this is only defined for PCA and MDS (aka PCoA) and not available for most other ordination methods.

Now, the problem is that this needs to be done manually and in a different way for PCA and MDS. It would be easier if the explained variance could be made more readily available for visualization and analysis purposes, for instance through a generic method that operates on the se object. This should be available for both PCA and MDS at least (possibly also for other covariance based methods that have this defined).

The plotting function could then automatically detect the availability of this information and to the plot as it is a recommended practice to show the explained variance when it is available. It could be even automatically (and optionally) used to tune the ratio between X and Y axis widths, which has been sometimes recommended.

14-microbiome-diversity.Rmd
```diff
@@ -106,15 +106,33 @@ between communities (samples).
...
+different way by different methods. [TODO add references and/or link
+to ordination chapter instead?]
+
+Some of the most common ordination methods in microbiome research
+include Principal Component Analysis (PCA), metric and non-metric
+multi-dimensional scaling (MDS, NMDS), The MDS methods is also known
+as Principal Coordinates Analysis (PCoA). Other recently popular
+techniques include t-SNE and UMAP. 
+
+Some of the above-mentioned ordination methods (PCA, MDS/PCoA) allow
```
FelixErnst, line 125: As far as I know, this only true for SVD/PCA and not for MDS,PCoA although `cmdscale` returns eigenvectors.

14-microbiome-diversity.Rmd
```diff
@@ -106,15 +106,33 @@ between communities (samples).
...
+axis; this is informative to show in the figures when it is defined
+and available.
+
 
 
 ### Estimating beta diversity
 
-In the following examples distances are calculated by variable
-functions supplied to the `FUN` argument. The function can defined by
+In the following examples dissimilarities are calculated by
```
FelixErnst, line 134: diversity/dissimilarities/distances are sometimes stand-ins for each other.

Can you add an explanation here how it is supposed to be understand in the context of `MiaBook`?

FelixErnst (approved):
Fine to me. If you want to add something to sections I commented on, let me know.

14-microbiome-diversity.Rmd
antagomir: This is commonly used with MDS/PCoA in microbiome research. 

It does not have an equally straightforward linear interpretation as PCA but it still seems to have [sensible interpretations](https://stats.stackexchange.com/questions/68680/how-to-interpret-variation-explained-by-principal-coordinates) although I must admit that I would still like to have a bit better intuition, too.

14-microbiome-diversity.Rmd
antagomir: Ah, this is quite essential difference technically since "distance" is a mathematically well-defined concept and many ecological beta diversity measures, such as Bray-Curtis, are not proper distances. It is more accurate to call them dissimilarities, and it is in a strict sense wrong to call them distances. It is not just semantics and I think we should aim to use technically correct vocabulary that is accepted also by the comp/stat oriented members of our community.

14-microbiome-diversity.Rmd
FelixErnst: `runPCA`/`calculatePCA` calculates the variance explained automatically. It is also included, if you plot `plotPCA`.

For `runMDS` it is different

14-microbiome-diversity.Rmd
FelixErnst: Thats a good explanation. Can you add this?

14-microbiome-diversity.Rmd
FelixErnst: If we can get a citation or a contribution from a statistician on this, this would be very helpful. The PR to scater I prepared added the support for the return of `eig` and `GOF` for `runMDS`. I refrained from calculating the variance, since I wasn't sure on it. In principle the same interfaces for `runPCA`/`plotPCA` can be used here as well.

(fyi: `runMDS2` will be gone for the release of `mia`)

FelixErnst:
@antagomir Do you want to add the additional explanation on difference/distance or should this be merged?

FelixErnst:
@antagomir should I merge this? See comment above for context

antagomir:
Working on it. Got buried under other running stuff.  I aim to close this tonight.

14-microbiome-diversity.Rmd
antagomir: Added, in the next update to the PR.

14-microbiome-diversity.Rmd
antagomir: I hope that I should have the required training (statistical ML) to pick up this one. We used to go through all this back in the days. But if someone notices any issues below I will be curious to learn more as this is a quick recap.

PCoA (Classical MDS) aims to minimize so-called "stress" ie. difference between pairwise similarities between the data points in the original (high-dimensional) vs. projected (low-dimensional) space. 

PCA is a special case of PCoA when we operate with Euclidean distances.

With Euclidean distances, we can minimize the stress by projecting to the maximal variance axes. Then the "proportion of explained variance" becomes a rather natural measure for how well the projected distances corresponds to the original distances.

With non-Euclidean distances, the PCoA uses a trick where the pointwise distances (or dissimilarities) are first cast into similarities a Euclidean space (with some information loss ie stress) and then projected to the maximal variance axes. In this case, the maximal variance axes do not any more directly measure the correspondence of the projected distances and original distances. Instead, they measure that correspondence in the already transformed space and interpretation is less straightforward. If the original space is sufficiently Euclidean, this may still be a good approximation (but then why not just use PCA). Explained variance percentages are still sometimes being used with PCoA. Most users probably never thought about it. 

We are probably more interested in measuring the correspondence between the projected and original distances i.e. the stress. Or relative stress, which is better comparable across different data sets: `sum((distEucSC - distMDS)^2)/sum(distEucSC^2)`. This [note](https://www.huber.embl.de/users/klaus/Teaching/statisticalMethods-lab.pdf) from Huber lab gives nice examples. 

Therefore, in fact, I would propose abandoning the idea of showing explained variance percentages for other than PCA ordination.

We could, however, add calculation of the stress and potentially the Shepard plot since these could be rather informative. These could be added either in the package directly, or in OMA. I am not sure what would be the best. Adding to OMA is more light weight at this point perhaps.

antagomir: Ah, I should be able to handle it per training. 

PCA is special case of PCoA with Euclidean distances. Then the variance maximizing procedure and explained percentages reflect directly the level of correspondence between the ordination and original space. With non-Euclidean distances the interpretation is different: the dissimilarities are first cast to Euclidean similarities with a trick, and only then the variance maximizing projections are calculated. This comes with information loss since non-Euclidean dissimilarities cannot, in general, be casted to Euclidean space without information loss. In practice, we are typically interested in quantifying the correspondence between the ordination and original space. We can do that directly by looking at the "stress" function instead. It measures that deviation. See [the note from Huber lab](https://www.huber.embl.de/users/klaus/Teaching/statisticalMethods-lab.pdf) for nice examples.  The explained variances for PCoA would measure how much variance is explained for the data that was already cast in the Euclidean space, and the link with the original data space distances is weaker.

In conclusion, I tend to not recommend showing explained variances for other than PCA. For PCoA, I add an example in OMA beta diversity chapter. We can later consider whether it should be made into its own function or not.

14-microbiome-diversity.Rmd
antagomir: Added.

PR #28, opened 2021-07-09
Adding content to the community typing section.

ChouaibB:
Transferred from course material the section about community typing, to the dedicated one at OMA; file: 15-microbiome-community.Rmd

### DIFF
--- a/15-microbiome-community.Rmd
+++ b/15-microbiome-community.Rmd
@@ -33,10 +33,11 @@ reflect the probability that a sample belongs to the given cluster.
 
 Let's cluster the data with DMM clustering. 
 
-```{r dmm, warning=FALSE}
+```{r dmm}
 # Runs model and calculates the most likely number of clusters from 1 to 7.
 # Since this is a large dataset it takes long computational time.
 # For this reason we use only a subset of the data; agglomerated by Phylum as a rank.
+tse <- relAbundanceCounts(tse)
 tse <- agglomerateByRank(tse, rank = "Phylum", agglomerateTree=TRUE)
 tse_dmn <- runDMN(tse, name = "DMN", k = 1:7)
 ```
--- a/15-microbiome-community.Rmd
+++ b/15-microbiome-community.Rmd
@@ -14,8 +14,117 @@ Heatmaps
 
 ## Community typing
 
-- Community State Types (CST)
-- Dirichlet Multinomial Mixtures (DMM)
+### Dirichlet Multinomial Mixtures (DMM)
+
+This section focus on DMM analysis. 
+
+One technique that allows to search for groups of samples that are
+similar to each other is the [Dirichlet-Multinomial Mixture
+Model](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0030126). In
+DMM, we first determine the number of clusters (k) that best fit the
+data (model evidence) using Laplace approximation. After fitting the
+model with k clusters, we obtain for each sample k probabilities that
+reflect the probability that a sample belongs to the given cluster.
+
+Let's cluster the data with DMM clustering. 
+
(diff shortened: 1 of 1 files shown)


15-microbiome-community.Rmd
```diff
@@ -14,8 +14,117 @@ Heatmaps
...
+
+Let's cluster the data with DMM clustering. 
+
+```{r dmm}
+# Runs model and calculates the most likely number of clusters from 1 to 7. 
+# For this small data, takes about 10 seconds. For larger data, can take much longer
+# because this demands lots of resources. 
+tse_dmn <- runDMN(tse, name = "DMN", k = 1:7)
+```
+
```
tvborm, line 37: The check has failed, if you open that you will see that `runDMN` function is not found, which means that `mia` package is not loaded.

I'm not completely sure how those pages are rendered, but there are other pages that has `library(mia)` or `mia::function()`.
 
It seems that those pages are rendered individually, so `mia `package needs to be loaded in every file. Is it working if you put `library(mia)` in the beginning of the file?

ChouaibB: That's true it does, I have checked by loading mia at begining but still another error will come up for variable tse missing.
And from se to tse variable at original code (at radboud) some cleaning and processing needs to be done before conversion (few extra chuncks of code need to be transfered as well). For this reason I kept the content raw because I was thinking there going to be later additions, that by then or at the end all will be re-rendered.

tvborm: OK, about the data thing: I think that we can't use the radboud data. However, you could use example data provided by `mia` package, e.g., `data(GlobalPatterns)`

ChouaibB: ok

antagomir: Yes, true. But keep the Rmd file that has full example with the course data because we like to have that available for course participants. I suggest that you 

1. modify this OMA workflow to use one of our open example data sets (e.g. GlobalPatterns)
2. add/keep the course data set Rmd in the Radboud material in github but hide it from the book menus, so that we can show it course participants as an example solution, after they have tried themselves to run community typing for the course data based on OMA example

Would this work?

ChouaibB: So you mean that we do the community typing and show the codes at OMA, but the recently removed chapter from radboud would be designed as an exercise using radboud dataset?

antagomir: Yes I mean that we 

1) do the community typing and show the codes at OMA using one of our open demo data sets (e.g. GlobalPatterns); should be easy to switch to another data; and 

2) the recently removed chapter from radboud is already designed to use the radboud dataset (unless I am mistaken) and it would be enough to simply copy the Rmd file in git, and make sure that it is not shown in the course material. It does not matter if it is available through the same github folder as long as it is not directly included in the course html material.

antagomir: I initially did not think the fact that we cannot use the Radboud data in OMA, hence I asked to move it as such. But these two changes should quick to implement.

ChouaibB: sure I'll do the changes

FelixErnst: Maybe also have a look at microbiomeDataSets for a dataset fitting best to the case/thought you want to get across with the section.

Regarding loading of the mia package: Have a look at the other `.Rmd` files in this book and also check for the things provided by the `rebook` package.

antagomir: However, do not pick the HITChip data from microbiomeDataSets (atlas1006, peerj, dietswap) as these are not sequencing studies and this may lead to confusion.

antagomir (approved):

PR #43, opened 2021-07-20
added separate alpha and beta diversity file

shigdel:
Here i have created separate rmarkdown for alpha and beta diversity section

### DIFF
--- a/alpha-diversity.rmd
+++ b/alpha-diversity.rmd
@@ -1,0 +1,154 @@
+---
+title: "alpha-diversity.rmd"
+output: html_document
+---
+
+```{r setup, echo=FALSE, results="asis"}
+library(rebook)
+chapterPreamble()
+```
+
+## Alpha diversity
+
+**_Alpha diversity_**, also sometimes interchangeably used with the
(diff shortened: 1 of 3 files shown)


community_comparisions.rmd
```diff
@@ -0,0 +1,121 @@
+## Community comparisons [TODO combine with the material above for simplicity?]
```
antagomir, line 1: fix the typo on the file name

antagomir: or in fact, I think this file can be just combined with the beta diversity file

alpha-diversity.rmd
```diff
@@ -0,0 +1,154 @@
+---
+title: "alpha-diversity.rmd"
+output: html_document
+---
+
+```{r setup, echo=FALSE, results="asis"}
+library(rebook)
+chapterPreamble()
+```
```
antagomir, line 9: These setups are not in the beta diversity file. Is there a reason? Or can these be included there as well?

antagomir (changes requested):
Also add modification to the file _bookdown.yml so that these changes become visible in the side bar.

PR #44, opened 2021-07-21
Adding composition barplot.

ChouaibB:
Hi,
In the following an example of composition barplot has been added to "Community composition" sub-section.
This is related to microbiome/miaViz#20 Composition barplot.


antagomir (approved):
Seems fine, will need to check the rendered output laetr.

Issue #15, opened 2021-03-01
LibrarySize


antagomir:
Refer to these commonly used concepts in MiaBook:

- `librarySizeFactors` is available from scuttle to calculate relative library sizes
- `colSums` is the function to get absolute values (total read counts per sample)

FelixErnst:
In addition `addPerFeatureQC` and `addPerCellQC` are available. These function attach summary data directly to the `rowData`/`colData`

antagomir:
Some more on this can be added to subsection titled "Library size".

PR #47, opened 2021-07-23
Added histogram example of library size visualization

TuomasBorman:
Hi,

I added a histogram example that visualizes library size distribution.

Issue: https://github.com/microbiome/OMA/issues/15

-Tuomas


antagomir (commented):
Seems good (although I could not check the visualizations).

I still think that this could deserve its own function (LibrarySize or ReadCount) but let us come back to that later.

PR #48, opened 2021-07-23
Alpha diversity boxplot example

TuomasBorman:
Hi,

here are an example of boxplot that visualizes Shannon index, and an example of pairwise Wilcoxon test.

-Tuomas

### DIFF
--- a/14a_alpha_diversity.Rmd
+++ b/14a_alpha_diversity.Rmd
@@ -111,25 +111,37 @@ head(colData(se)$shannon)
 
 Alpha diversities can be visualized with boxplot. Here, Shannon index is compared 
 between different sample type groups. Individual data points are visualized by 
-plotting them as points. With more data points, test `geom_jitter` function
-instead of `geom_point`.
+plotting them as points with `geom_jitter`.
+
+`geom_signif` is used to test, if these differences are statistically significant.
+It adds p-values to plot.
 
 ```{r visualize-shannon}
+if( !require(ggsignif) ){
+  install.packages(ggsignif)
+}
 library(ggplot2)
-ggplot(as.data.frame(colData(se)), aes(x = SampleType, y = shannon)) +
+library(ggsignif)
+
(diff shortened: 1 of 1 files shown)


14a_alpha_diversity.Rmd
```diff
@@ -109,6 +109,29 @@ se <- mia::estimateDiversity(se,
...
+plotting them as points. With more data points, test `geom_jitter` function
+instead of `geom_point`.
+
+```{r visualize-shannon}
+library(ggplot2)
+ggplot(as.data.frame(colData(se)), aes(x = SampleType, y = shannon)) +
+  # Outliers are removed, because otherwise each data point would be plotted twice; 
+  # as an outlier of boxplot and as a point of dotplot.
+  geom_boxplot(outlier.shape = NA) + 
+  geom_point() +
```
antagomir, line 123: wouldn't `geom_jitter` work better? The jitter width could be set to 0.5 or something suitable.

14a_alpha_diversity.Rmd
```diff
@@ -109,6 +109,29 @@ se <- mia::estimateDiversity(se,
...
+```
+To test, if these differences that can be seen in plot are statistically significant
+we do pairwise Wilcoxon test. It compares every group with each others to see if 
+their values differ. 
+
+```{r pairwise-wilcoxon}
+df <- as.data.frame(colData(se))
+pairwise.wilcox.test(df$shannon, df$SampleType,  p.adjust.method = "fdr")
+```
+
```
antagomir, line 134: Would be ideal to show the differnces in the figure itself using `ggsignif` or other suitable packages?

14a_alpha_diversity.Rmd
TuomasBorman: Changed to jitter

14a_alpha_diversity.Rmd
TuomasBorman: Removed wilcoxon and added geom_signif

antagomir:
Thanks

PR #50, opened 2021-07-23
3D interactive plot

ChouaibB:
Hi,

Added is an example of 3D interactive plot using plotly.
Worked also when rendered to html.


antagomir (changes requested):
- Can you add a simple link to this in beta diversity section
- No dashes in file names; replace "-" with "_"

antagomir (approved):

antagomir:
@FelixErnst this changes all dashes to underscores in file names. Just cross-checking it is ok.

TuomasBorman (approved):

PR #51, opened 2021-07-24
Deleted duplicated files

TuomasBorman:
Hi,

due to renaming of files (PR: https://github.com/microbiome/OMA/pull/50), `ghpages` branch includes currently duplicated files; one with old name and one with new name. This removes those files and directories that have duplicated copies, and leaves ones with new names. 

-Tuomas


antagomir:
Seems fine.

Issue #34, opened 2021-07-15
Separate alpha and beta diversity in their own sections.


antagomir:
These are so different concepts that it is better to treat them in two separate sections.

antagomir:
I propose that @shigdel implements this and @ChouaibB is available to help in the process.

Issue #52, opened 2021-07-24
Keeping the book clean


FelixErnst:
Hi 

I noticed a lot of pecuilar things, which will be very difficult to fix later on.

1. There are two `Rmd` files which begin with `14`, `14a` and `14b`. This is not good and does not follow the setup of the book.
2. 
![image](https://user-images.githubusercontent.com/17824332/126875710-f575aafd-0dae-46b4-b8a4-9d8c69f661b5.png)

I don't know, where the second session info comes from in the middle, but it not helpful nor necessary.

3. There seems to be a plotting section after the bibliography. I think that doesn't make sense
![image](https://user-images.githubusercontent.com/17824332/126875787-fd0eded1-6611-4b1a-a7d4-8871055f14d2.png)

I think these needs to be cleaned up, before the book becomes unmaintainable and unreadable.

antagomir:
Yes. @ChouaibB can you make a PR to fix the obvious errors overlined above (sections in wrong places)

The 14a/14b can be changed into 14 & 15; and other numbers can be bumped up correspondingly.. we will have to work out the numbering scheme in more detail but this is to fix the immediate issues.

You can run in R bookdown::serve_book()

antagomir:
Hmm the "Interactive 3D plots" is still hanging around after Chapter 13. @ChouaibB can you remove that from the side panel?

ChouaibB:
I gave it a try to have it as a hidden page that can be rendered with the automatic flow of github so as it is available as html page when referred to at [the section](https://microbiome.github.io/OMA/beta-diversity.html#visualizing-the-most-dominant-genus-on-pcoa), but unfortunately I did not figure it out yet.
@TuomasBorman suggested to have an Extras chapter [at the discussion](https://github.com/microbiome/OMA/pull/54).
I can try solving it still.

antagomir:
Thanks - we need to find some way to include extra material without adding it to the side bar. An intermediate solution can be to have an Appendix chapter.

ChouaibB:
ok, I add another Appendix dedicated to Extras.

PR #54, opened 2021-07-26
Book fixes and clean-up.

ChouaibB:
In the following fixes were made related to issue #52.
* Files renamed
* Fixing alpha and beta sections as were supposed to be separate.
* Added session info to alpha diversity section.
* Moved the interactive 3D plot section before bibliography.


TuomasBorman:
Thanks! Overall, I think the book looks nice. 

One thing: Should `Interactive 3D plots` be a subchapter of e.g., `Extras` chapter? If there will be more these extra stuffs, then again, I think the structure of book is not that coherent anymore. Any thoughts? @FelixErnst 

image
image
image
image

TuomasBorman:
And some other things: 

1. Should `Alpha diversity` and `Beta diversity` be subchapters of `Microbiome exploration`?
2. Some headers are written so that every word starts with uppercase, like `8 Microbiome Community`, some aren't, e.g., `10 Microbiome time series`. This should be consistent

ChouaibB:
1. the separation was related to the closed issue: https://github.com/microbiome/OMA/issues/34
2. Yes I can fix that.

antagomir (commented):
The extra material should be only linked in text and not included in side bar. I had asked so in the original comments but good to reiterate. There will be more of such complementary material.

Alpha and beta diversity are vast and very different topics and both deserve their own section. Often they are also used in exploration pääse, we can cite these sections there.

I will have a look at the overall structure in more detalji when getting out of the forests

antagomir (approved):

Issue #27, opened 2021-07-08
Rarity examples


antagomir:
Add example/s and explanation of rarity indices in alpha diversity section.

PR #58, opened 2021-07-28
Added rarity index example

TuomasBorman:
Hi,

here is an example of rarity index.

Issue: https://github.com/microbiome/OMA/issues/27

-Tuomas


antagomir (approved):

TuomasBorman:
> This can also be included in the figure below as we have an uneven no. of panels at the moment.
> 
> ```
> lapply(c("observed", "shannon","simpson", "relative", "faith","log_modulo_skewness"),
>                 plotColData,
>                 object = se,
>                 x = "SampleType",
>                 colour_by = "SampleType")
> ```

True, looks better now. I added it directly to master

PR #62, opened 2021-08-03
Bookstyle change

### LINKED ISSUE #21: Book style
We could switch to using the [msmbstyle](https://github.com/grimbough/msmbstyle) that provides a [nice output](https://www-huber.embl.de/users/msmith/msmbstyle/figures-margin-content.html) and was used in the [Huber&Holmes MSMB book](https://www-huber.embl.de/msmb/).

TuomasBorman:
Hi,

this changes the book style to `msmbstyle`. You can render it, e.g., with `bookdown::serve_book()`
This is not completely working yet, for instance, there is problem with table of contents. I wanted to hear you opinion before I use more time to fix this thing. 

Personally, I think that current book style is better. It is more clear and user can easily find what he/she is looking for because toc is always visible. In addition to that, I think that `msmbstyle` doesn't have those "download pdf" and "share" icons (currently we don't have pdf version, but we could easily add it).

Any thoughts?

Issue: https://github.com/microbiome/OMA/issues/21

-Tuomas

### DIFF
--- a/01_introduction.Rmd
+++ b/01_introduction.Rmd
@@ -1,4 +1,4 @@
-# (PART) Introduction {-}
+
 
 # Data Infrastructure {#data-introduction}
 
--- a/10_overview.Rmd
+++ b/10_overview.Rmd
@@ -1,4 +1,3 @@
-# (PART) Focus Topics {-}
 
 # Overview
 
--- a/90_example_data.Rmd
+++ b/90_example_data.Rmd
@@ -1,4 +1,4 @@
-# (PART) Appendix {-}
+
 
 # Example Data {#example-data}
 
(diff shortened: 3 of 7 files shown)


antagomir:
Thanks a lot. 

Assuming that the navigation would work, the layout seems a bit nicer and more professional to me in the `msmbstyle` (for which we have a properly functioning example in [Huber & Holmes](https://www-huber.embl.de/users/msmith/msmbstyle/)) than in [standard bookdown](https://microbiome.github.io/OMA/)

But I can accept either format easily. 

Would be therefore good to hear opinions from others, too.

ChouaibB:
I have checked the result of the rendering, and surely I guess the features at the older style e.g. pdf download, icon for copying code snippets could be added and all.
However, I personally still like the original layout. Thanks.

Issue #63, opened 2021-08-03
Front page figure


antagomir:
Copy the [front page image from Radboud summer school page](https://microbiome.github.io/course_2021_radboud/) to OMA front page (including citation)?

PR #64, opened 2021-08-05
Added front page figure

TuomasBorman:
Hi,

I like Leo's idea of front page figure so I implemented it. 

Here are two different versions, another has figure before title and another after. Which is better? I think after is better, and in this current implementation it is after. 

Any thoughts from others, should we add this?

Before:
image

After (current implementation):
image

Issue: https://github.com/microbiome/OMA/issues/63

-Tuomas

### DIFF
--- a/index.Rmd
+++ b/index.Rmd
@@ -49,3 +49,17 @@ knitr::opts_chunk$set(
   dpi=300
 )
 ```
+
+```{js, echo = FALSE}
+// This block adds image to the front page
+title=document.getElementById('header');
+title.innerHTML = title.innerHTML + 
+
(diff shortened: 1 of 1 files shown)


antagomir:
I also vote for the latter. Thanks!

antagomir:
Hmm just one note - would it be possible to remove the ML4microbiome logo from the top left corner e.g. with an image manipulation program? We are associated with that COST action but I would not tie the whole book with it.

TuomasBorman:
OK, I can do that

ChouaibB:
After seems perfect.

TuomasBorman:
Removed 

image

antagomir:
so nice

ChouaibB (approved):

PR #67, opened 2021-08-06
Diffab

HenrikEckermann:
first draft

### DIFF
--- a/.github/workflows/build-book-release.yml
+++ b/.github/workflows/build-book-release.yml
@@ -151,6 +151,12 @@ jobs:
           remotes::install_cran("rcmdcheck")
           BiocManager::install("BiocCheck")
         shell: Rscript {0}
+        
+      - name: Install tidyverse
+        run: |
+          message(paste('****', Sys.time(), 'installing tidyverse ****'))
+          install.packages("tidyverse")
+        shell: Rscript {0}
 
       - name: Session info
         run: |
--- a/.github/workflows/build-book-release.yml
+++ b/.github/workflows/build-book-release.yml
@@ -157,7 +157,31 @@ jobs:
           message(paste('****', Sys.time(), 'installing tidyverse ****'))
           install.packages("tidyverse")
         shell: Rscript {0}
-
+        
+      - name: Install tidySummarizedExperiment
+        run: |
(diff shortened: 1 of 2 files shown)


HenrikEckermann:
The above error seems to arise because tidyverse is not available on the server. Potentially also other packages I used are not there, e.g. patchwork and tidySummarizedExperiment. Is that a problem or how can I fix this?

antagomir:
I think you can add the installation command for those packages in the GHA (Github Action) workflow in
.github/workflows/check-bioc-devel.yml (or the other yml file in that same dir).

HenrikEckermann:
I would not need the whole tidyverse collection for this file but as it is now I use these 4:  

```r
library(dplyr) 
library(tibble)
library(purrr)
library(ggplot2)
```


Only purrr seems to be missing when I looked at the two files you mentioned. I typically just load the tidyverse. I could change that to loading the single packages if that is preferred. I will await response before I will change the files

antagomir:
I do not see a problem in including the whole tidyverse. I think it has to  be installed just once on a given server, so it would not add too much. On the other hand, if we do not need, perhaps it would be easier and faster (and more CO2 friendly) to add just those we need?

antagomir:
I do not see a problem in including the whole tidyverse. I think it has to  be installed just once on a given server, so it would not add too much. On the other hand, if we do not need, perhaps it would be easier and faster (and more CO2 friendly) to add just those we need?

HenrikEckermann:
I edited `.github/workflows/build-book-release.yml` and added tidyverse, patchwork, DESeq2 and ANCOMBC. Now the checks passed. Please merge in case this is how you would like to add the packages. I did not edit anything in `DESCRIPTION`.  So, if that was preferred, we need to change it.

HenrikEckermann:
Just saw that other used this approach:

```r
if (!require(package)){
  # install the package
  install.packages(package)
}
```

In that case I do not need to edit the yaml or DESCRIPTION file. If that is preferred then, I will undo the changes to the yaml and include install code in the diffab section iteself...

antagomir:
Ok with me. We change later ifn needed. 

Might be good to check what they did with OSCA book.

Issue #61, opened 2021-08-03
Clustering examples


antagomir:
Aaron Lun has done some nice work on integrating clustering methods with SE class in [bluster](https://www.bioconductor.org/packages/devel/bioc/vignettes/bluster/inst/doc/clusterRows.html). Some examples in OMA could be useful.

ChouaibB:
Would adding an example of [graph-based-clustering](https://www.bioconductor.org/packages/devel/bioc/vignettes/bluster/inst/doc/clusterRows.html#graph-based-clustering) for instance using louvain algorithm at [community-typing](https://microbiome.github.io/OMA/microbiome-community.html#community-typing) be a good place? e.g a subsection titled "Community Detection"

ChouaibB:
When I tried the graph based clustering with PCoA done at the [community-typing](https://microbiome.github.io/OMA/microbiome-community.html#community-typing) section, the following communities were obtained (short-walk vs louvain algorithm):
![image](https://user-images.githubusercontent.com/86964562/128693307-1276d77a-0795-4bfe-8d53-fd3f7397c75f.png)
![image](https://user-images.githubusercontent.com/86964562/128693337-34f17b70-1595-445e-9173-dbfcb19b46c1.png)

PR #71, opened 2021-08-09
Adding draft example.

ChouaibB:
This is a draft example of using the package  [bluster](https://www.bioconductor.org/packages/devel/bioc/vignettes/bluster/inst/doc/clusterRows.html), graph based clustering for community discovery.
Related issue #61 .

### DIFF
--- a/16_microbiome_community.Rmd
+++ b/16_microbiome_community.Rmd
@@ -229,20 +229,22 @@ if(!require(bluster)){
 
 The algorithm used is "short random walks" [@Pons2006]. 
 Graph is constructed using different k values (the number of nearest neighbors to consider during graph construction) 
-using the clr assay data. Then plotting the communities using UMAP [@McInnes2018] ordination as a visual exploration aid.
+using the robust centered log ratio (rclr) assay data. Then plotting the communities using UMAP [@McInnes2018] ordination as a visual exploration aid.
 
 ```{r, message=FALSE, warning=FALSE}
 library(bluster)
 library(patchwork) # For arranging several plots as a grid
 library(scater)
 
+tse <- transformCounts(tse, method = "rclr")
+
 # Performing and storing UMAP
-tse <- runUMAP(tse, name="UMAP", exprs_values="clr")
+tse <- runUMAP(tse, name="UMAP", exprs_values="rclr")
 
 k <- c(2,3,5,10)
 ClustAndPlot <- function(x) {
   # Creating the graph and running the short random walks algorithm  
-  graph_clusters <- clusterRows(t(assays(tse)$clr), NNGraphParam(k=x))
+  graph_clusters <- clusterRows(t(assays(tse)$rclr), NNGraphParam(k=x))
   
   # Results of the clustering as a color for each sample
   plotUMAP(tse, colour_by = I(graph_clusters)) +
@@ -265,10 +267,10 @@ for each case presented earlier. For more about Silhouettes read [@Rousseeuw1987
 
 ClustDiagPlot <- function(x) {
   # Getting the clustering results
-  graph_clusters <- clusterRows(t(assays(tse)$clr), NNGraphParam(k=x))
+  graph_clusters <- clusterRows(t(assays(tse)$rclr), NNGraphParam(k=x))
   
   # Computing the diagnostic info
-  sil <- approxSilhouette(t(assays(tse)$clr), graph_clusters)
+  sil <- approxSilhouette(t(assays(tse)$rclr), graph_clusters)
   
   # Plotting as a boxlpot to observe cluster separation
   boxplot(split(sil$width, graph_clusters), main=paste0("k = ", x))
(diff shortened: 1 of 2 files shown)


antagomir (changes requested):
Thanks. 

Now checking the approach in more detail I see that the clustering is done based on the reduced dimensionality (2D). This is suitable to characterize easily the 2D plots but not for actual high-dimensional community typing because ordination to 2D looses considerable amount of information (often ~80% when we talk about human gut microbiome for instance). 

Hence I would refer to this as a visual exploration aid, and just show a single example (exlucde the ClustAndPlot2 function to keep it simple).

16_microbiome_community.Rmd
```diff
@@ -210,6 +210,52 @@ euclidean_dmm_plot <- ggplot(data = euclidean_dmm_pcoa_df,
...
+euclidean_pcoa_df$communities <- graph_clusters
+# Results of the clustering as a color for each sample  
+ggplot(data = euclidean_pcoa_df, 
+                             aes(x=pcoa1, y=pcoa2,
+                                 color = communities)) +
+  geom_point() +
+  labs(x = "PC1",
+       y = "PC2",
+       title = paste0("k = ", x))
+}
```
TuomasBorman, line 251: Is the indentation little bit weird here? First 2 lines of function are correct, but all other do not have indentation

ChouaibB: yes it is weird, I fix it.

TuomasBorman (approved):

16_microbiome_community.Rmd
```diff
@@ -229,22 +229,26 @@ if(!require(bluster)){
...
   graph_clusters <- clusterRows(t(assays(tse)$clr), NNGraphParam(k=x))
-  euclidean_pcoa_df$communities <- graph_clusters
   
-  # Results of the clustering as a color for each sample  
-  ggplot(data = euclidean_pcoa_df, aes(x=pcoa1, y=pcoa2, color = communities)) +
-    geom_point() +
-    labs(x = "PC1", y = "PC2", title = paste0("k = ", x))
+  # Results of the clustering as a color for each sample
+  plotUMAP(tse, colour_by = I(graph_clusters)) +
+    labs(title = paste0("k = ", x))
```
antagomir: I find it somehow confusing that the ordination (UMAP) and clustering (NNGraph) are done in different domains (log10 vs. clr). Does UMAP strictly require log10p transformation, or could it be operate with clr as well?  It would be better for consistency to have both ordination & clustering operating on the same data.

16_microbiome_community.Rmd
```diff
@@ -254,6 +258,28 @@ plots <- lapply(k,ClustAndPlot)
 (plots[[1]] + plots[[2]]) / (plots[[3]] + plots[[4]])
 ```
 
+Similarly the _`bluster`_ [@R-bluster] package offers clustering diagnostics
+that can be used for judging the clustering quality (see [Assorted clustering diagnostics](http://bioconductor.org/packages/release/bioc/vignettes/bluster/inst/doc/diagnostics.html)).
+In the following, silhouette width as a diagnostic tool is computed and results are visualized
```
antagomir, line 263: I think it should "Silhouette" (capitalized)

antagomir: Can you add citation to the original Silhouette index publication if it can be found?

16_microbiome_community.Rmd
```diff
@@ -229,22 +229,26 @@ if(!require(bluster)){
 
 The algorithm used is "short random walks"[@Pons2006]. 
 Graph is constructed using different k values (the number of nearest neighbors to consider during graph construction) 
-using the clr assay data. Then plotting the communities using the earlier PCoA ordination as a visual exploration aid.
+using the clr assay data. Then plotting the communities using UMAP ordination as a visual exploration aid.
```
antagomir, line 232: Add (original) reference to UMAP?

16_microbiome_community.Rmd
```diff
@@ -229,22 +229,26 @@ if(!require(bluster)){
 
 The algorithm used is "short random walks"[@Pons2006].
```
antagomir, line 230: space missing after the second quote?

antagomir (changes requested):

16_microbiome_community.Rmd
ChouaibB: At first look at documentation I thought it was strictly logcounts. However it is not, can be done with clr as well. I change it.

16_microbiome_community.Rmd
ChouaibB: The documentation of `approxSilhouette` does not indicate the publication, only an additional explanation of the approximation; at the source code file. However, there is the following [publication](https://www.sciencedirect.com/science/article/pii/0377042787901257?via%3Dihub), shall I refer to it?

ChouaibB: The exact[ Silhouette calculation at cluster package](https://www.rdocumentation.org/packages/cluster/versions/2.1.2/topics/silhouette) (referred to by bluster) is using as a reference the same publication above.

16_microbiome_community.Rmd
antagomir: Now as method "rclr" has been implemented perhaps you could use that?

16_microbiome_community.Rmd
antagomir: ok

antagomir:
ok after the requested changes

PR #74, opened 2021-08-11
integrated fido with current diffab section

HenrikEckermann:
first integration of what I already had for diffab and the section about `fido`. The `fido` section works with the `SprockettTHData()`. I tried a subset of that first time I tried to write the diffab section but ran into an uncommon error when using DESeq2. I will check this out one more time at a later time point. Currently `fido` section is also at phylum level and the diffab section is at genus level.

### DIFF
--- a/17_differential_abundance.Rmd
+++ b/17_differential_abundance.Rmd
@@ -6,13 +6,312 @@ chapterPreamble()
 ```
 
 
+## Introduction to differential abundance analysis (DAA)
 
-A number of methods for differential abundance analysis are available,
-and reviewed elsewhere.
+*What is DAA and why do we do it?*
+
+In this section we will work with one of the [openly available datasets](https://rdrr.io/github/FelixErnst/mia/man/mia-datasets.html)in mia to illustrate how to perform differential abundance analysis (DAA). DAA aims to identify differences in taxon abundance between two or more groups (e.g. treatment vs control) for each taxon of a sample. This can be performed at any phylogenetic level but we will work at Genus level abundances in this section. We perform DAA to identify biomarkers and/or gain understanding of a complex system by looking at its isolated components. For example, identifying that a bacterial taxon is different between e.g. a patient group with disease *X* vs a healthy control group might lead to important insights into the pathophysiology. Changes in the microbiota might be causal or a consequence of the disease. Either way, it can help to understand the system as a whole. Be aware that this approach has also been critisized [recently](https://arxiv.org/abs/2104.07266).
+
+
+
+*How to do DAA*
+
+There are many tools to perform DAA. The most popular tools, without going into evaluating whether or not they perform well for this task, are:  
+- [ALDEx2](https://bioconductor.org/packages/release/bioc/html/ALDEx2.html)   
+- [ANCOM](https://github.com/FrederickHuangLin/ANCOM) [(BC)](https://bioconductor.org/packages/release/bioc/html/ANCOMBC.html)  
+- [corncob](https://cran.r-project.org/web/packages/corncob/index.html)  
+- [DESeq2](https://bioconductor.org/packages/release/bioc/html/DESeq2.html)  
+- [edge](https://bioconductor.org/packages/release/bioc/html/edgeR.html)  
+- [LEFse](https://bioconductor.org/packages/release/bioc/html/lefser.html)  
+- [MaAsLin2](https://www.bioconductor.org/packages/release/bioc/html/Maaslin2.html)  
+- [metagenomeSeq](https://www.bioconductor.org/packages/release/bioc/html/metagenomeSeq.html)  
+- [limma voom](https://bioconductor.org/packages/release/bioc/html/limma.html)  
+- [t-test](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/t.test)  
+- [Wilcoxon test](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/wilcox.test)  
+
+
+We recommend you to have a look at [Nearing et al.](https://www.biorxiv.org/content/10.1101/2021.05.10.443486v1.full) who compared all these listed methods across 38 different datasets. Because different methods have different approaches (parametric vs non-parametric, different normalization techiniques etc.) to perform the same task (test differential abundance), results can differ between methods. Unfortunately, as Nearing et al. point out, they differ disturbingly much. Therefore, it is highly recommended to pick several methods to get an idea about how robust and potentially reproducible your findings are depending on the method. In this section we demonstrate 3 methods (ANCOM-BC, Wilcoxon-test and Deseq2) and will compare the results between them. Note that the purpose of this section is to show how to perform DAA in R, not how to correctly do causal inference. E.g. there might be confounding factors that might drive (the absence of) differences between the shown groups that we ignore for simplicity. Furthermore, we picked a dataset that merely has microbial abundances in a TSE object as well as a grouping variable in the sample data. We simplify the analysis by only including 2 of the 3 groups. 
+
+
+```{r load-pkg-data}
+# packages we use in this tutorial and whether they will be installed via 
+# Bioconductor.
+required_pkg <- list(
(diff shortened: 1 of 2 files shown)


17_differential_abundance.Rmd
```diff
@@ -6,13 +6,312 @@ chapterPreamble()
 ```
 
 
+## Introduction to differential abundance analysis (DAA)
+
+*What is DAA and why do we do it?*
+
+In this section we will work with one of the [openly available datasets](https://rdrr.io/github/FelixErnst/mia/man/mia-datasets.html)in mia to illustrate how to perform differential abundance analysis (DAA). DAA aims to identify differences in taxon abundance between two or more groups (e.g. treatment vs control) for each taxon of a sample. This can be performed at any phylogenetic level but we will work at Genus level abundances in this section. We perform DAA to identify biomarkers and/or gain understanding of a complex system by looking at its isolated components. For example, identifying that a bacterial taxon is different between e.g. a patient group with disease *X* vs a healthy control group might lead to important insights into the pathophysiology. Changes in the microbiota might be causal or a consequence of the disease. Either way, it can help to understand the system as a whole. Be aware that this approach has also been critisized [recently](https://arxiv.org/abs/2104.07266).
```
TuomasBorman, line 13: 1) This website... Should it be https://microbiome.github.io/mia/reference/mia-datasets.html

2) ia/man/mia-datasets.html)in mia to illusrate...
There is no space between web address and "in"

17_differential_abundance.Rmd
```diff
@@ -6,13 +6,312 @@ chapterPreamble()
...
... have a look at [Nearing et al.](https://www.biorxiv.org/content/10.1101/2021.05.10.443486v1.full) who compared all these listed methods across 38 different datasets. Because different methods have different approaches (parametric vs non-parametric, different normalization techiniques etc.) to perform the same task (test differential abundance), results can differ between methods. Unfortunately, as Nearing et al. point out, they differ disturbingly much. Therefore, it is highly recommended to pick several methods to get an idea about how robust and potentially reproducible your findings are depending on the method. In this section we demonstrate 3 methods (ANCOM-BC, Wilcoxon-test and Deseq2) and will compare the results between them. Note that the purpose of this section is to show how to perform DAA in R, not how to correctly do causal inference. E.g. there might be confounding factors that might drive (the absence of) differences between the shown groups that we ignore for simplicity. Furthermore, we picked a dataset that merely has microbial abundances in a TSE object as well as a grouping variable in the sample data. We simplify the analysis by only including 2 of the 3 groups.
```
TuomasBorman, line 33: Can you split this text to own separate lines?

TuomasBorman: Deseq2 --> DESeq2

17_differential_abundance.Rmd
```diff
@@ -6,13 +6,312 @@ chapterPreamble()
...
+# check the number of differentially abundant taxa between groups
+ggplot(wilcoxon_p, aes(p_adjusted)) +
+  geom_histogram()
+```
+
+
+
+
+### DESeq2
+Our second analysis method is DESeq2. DESeq2 utilizes the negative binomial distribution to detect differences in read counts between groups. It performs a data normalization automatically that accounts for differences between library sizes. It will also automatically output adjusted p-values. More information about this method can be found, e.g., from Harvard Chan Bioinformatic Core's
```
TuomasBorman, line 116: Can you split this text to own lines?

17_differential_abundance.Rmd
TuomasBorman: And other too long lines also...? You can check other files, how wide the line should be

TuomasBorman (changes requested):
I think this looks nice, some minor things only

17_differential_abundance.Rmd
```diff
@@ -6,20 +6,411 @@ chapterPreamble()
...
+
+We see that while each method at least identified 14 genera as differentially
+abundant, they only agree on 5 genera. Lets plot those 5 genera:
+
+
+```{r}
+# create a plot for each genus where the score is indicated in the title
+plots <- pmap(select(summ, genus, score), function(genus, score) {
+  ggplot(daa, aes_string("x", genus)) +
+  geom_boxplot() +
```
antagomir, line 313: How about adding `geom_jitter(width=0.2, alpha=0.5)` or similar? This can potentially uncover interesting differences that some methods can/not take into account.

17_differential_abundance.Rmd
```diff
@@ -6,20 +6,411 @@ chapterPreamble()
...
+)
+
+
+# install required packages
+for (pkg in names(required_pkg)) {
+  if (required_pkg[[pkg]] == "BIOC" & !pkg %in% installed.packages()) {
+    BiocManager::install(pkg)
+  } else if (required_pkg[[pkg]] == "CRAN" & !pkg %in% installed.packages()) {
+    install.packages(pkg, dependencies = TRUE)
+  }
```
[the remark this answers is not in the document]
antagomir: If we like to show how to do it when necessary, it could be in its own file and just a link would be included in the text?

17_differential_abundance.Rmd
```diff
@@ -39,7 +430,9 @@ tse <- SprockettTHData()
 names(colData(tse))
```
antagomir, line 430: I would hide this - secondary and takes lot of space.

antagomir: Same for line 441 and other similar lines. In most cases we do not need to show the assay. Let us try to highlight the most relevant content to understand the main features of the method, and hide other material in the background

antagomir (commented):
Thanks! 

Some remaining comments can be found in this review.

The visualization of prior distributions in fido could be hidden as potentially confusing. 

Overall, very nice. We can think if the chapter can be somehow simplified further by

- hiding (not showing) some secondary material
- showing just 2 (not 3) DAA methods that we think that would actually deserve to be widely used as default choice (I am not sure if either  Wilcoxon or DESeq2 is such method - although I am currently indecisive which methods to use). However let us not discard any examples that have been written, we can always create supplementary files that are not shown in the text but are behind a link as R code for instance.

17_differential_abundance.Rmd
```diff
@@ -6,20 +6,411 @@ chapterPreamble()
...
+
+# format data so that we can perform the wilcoxon tests using map 
+x <- "pheno"
+daa <- assay(tse, "clr") %>% 
+          t() %>%
+          data.frame() %>%
+          rownames_to_column("Sample_ID") %>%
+          add_column(x = colData(tse)[[x]])
+# select all genera that we perform the test on
+genera <- select(daa, -Sample_ID, -x) %>% colnames()
```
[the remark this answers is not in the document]
HenrikEckermann: we use the `daa` dataframe in the `map_dfr` body below and the `genera` vector is created as we only "loop" through the genera, not also `Sample_ID` and `x`. I will add comments to clarify.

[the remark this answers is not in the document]
antagomir: I was just thinking if the code could be simplified and picking genus data straight from the tse object would seem more straightforward. Although potentially more error prone too if it has to be linked with these columns anyway. The faster the examples can be grasped the better.

HenrikEckermann:
Now, all feedback points (here and in slack) should have been addressed. Let me know if you see any remaining issues @TuomasBorman @antagomir

TuomasBorman:
Now it looks nice

1. I'm not sure about that output folder of `Maaslin2`, I think it does not cause problems to the published book itself

2. Github hasn't ran those checks. That might be because of one conflict. I think after resolving it we might see if that extra folder by `Maaslin2` cause problems

antagomir:
Hi - conflicts must be resolved still but seems very useful overall.

If maaslin folder is a problem, we could do eval=FALSE in chunk header in this exceptional case

HenrikEckermann:
I cannot see what the exact conflicts are if I use the code that git suggests for a command line merge. I see there are a lot o changes in the master. I will create a new branch and pull request where I started from this master. I dont want to mess anything up...

PR #78, opened 2021-08-17
Reorganizing some chapters and fixed minor bugs

antagomir:


antagomir:
Also initiated network chapter.

PR #89, opened 2021-09-15
Add speed comparisons to Extra

RiboRings:
Hi! I finished up the plots on speed comparisons and resolved the csv data set. I also added some explanatory text, which I intend to extend a little bit more. I am planning to include the Grieneisen and Hintikka data sets in the experiments in the future when they get available.


antagomir (commented):
Did you try to run  in R:
`bookdown::serve_book()`

This will allow you to browse to book and view your changes as a html.

I had to make two changes to get that working, I have no pushed them to this speed_comparisons branch.

The file "execution_times.csv" is called in the Rmd file. Where does this come from, and do we really need it?

Can we separate the speed comparisons to a separate subchapter (out from 97_interactive_3d_plots.Rmd); just create another Rmd file with a different name, use similar header for the file than in the other Rmd files (modify as needed), and list the Rmd file it in _bookdown.yml - then the new chapter should appear when running bookdown::serve_book()

Issue #104, opened 2021-10-27
Example of cross correlation


TuomasBorman:
Hi,

as Leo suggested, OMA should include example of using `getExperimentCrossCorrelation`. Where it should be added? Should there be own chapter for cross correlation or more broadly about multiomics? I think maybe the latter...

Naming of the chapter: `Cross correlation`, `Multiomics`...?

[My poster example](https://github.com/TuomasBorman/MIHD2021_example_workflow_for_poster/blob/main/example_workflow.pdf) might have something that could be used?

-Tuomas

PR #105, opened 2021-10-28
Cross-correlation example to OMA

TuomasBorman:
Hi,

there was already a file that included cross-correlation example. I modified it and changed the data. Also, it was not included in OMA, but now there is a chapter dedicated to multi-assay analyses.

I'm not sure if that text about importing should be added. 

Issue: https://github.com/microbiome/OMA/issues/104

-Tuomas

### DIFF
--- a/23_multi-assay_analyses.Rmd
+++ b/23_multi-assay_analyses.Rmd
@@ -96,14 +96,14 @@ mae[[1]] <- agglomerateByRank(mae[[1]], rank = "Genus")
 mae[[1]] <- transformSamples(mae[[1]], method = "log10", pseudocount = 1)
 
 # Cross correlates data sets
-correlation_table <- testForExperimentCrossCorrelation(mae, 
-                                                       experiment1 = 1,
-                                                       experiment2 = 2,
-                                                       abund_values1 = "log10", 
-                                                       abund_values2 = "nmr",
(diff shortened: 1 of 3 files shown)


23_multi-assay_analyses.Rmd
```diff
@@ -0,0 +1,169 @@
...
+"If this bacteria is present, is this metabolite's concentration then low or high"?
+
+```{r cross-correlation5}
+# Agglomerate microbiome data at Genus level
+mae[[1]] <- agglomerateByRank(mae[[1]], rank = "Genus")
+# Does log10 transform for microbiome data
+mae[[1]] <- transformSamples(mae[[1]], method = "log10", pseudocount = 1)
+
+# Cross correlates data sets
+correlation_table <- testForExperimentCrossCorrelation(mae,
```
antagomir, line 99: Just started to think if we could/should have a shorter name like `testExperimentCrossCorrelation`.. is the "For" important?

antagomir (approved):
looks good!

23_multi-assay_analyses.Rmd
TuomasBorman: shorter name might be better indeed, I have some other improvement ideas also so I can make a PR where we can discuss about these

PR #116, opened 2021-12-20
Bicluster

TuomasBorman:
Hi,

this PR includes biclustering example.

1. Should I use `cobiclust` R pkg instead of `biclust`?
2. Should `biclust` be done with `clr` abundances?

3.  This heatmap includes annotations which are done just by using `ggplot2`. 
   - Point of this is to show how to do this without using additional packages. 
   - However, the point of this is biclustering and not different ways to do heatmap, and this might take focus from the topic
   - Should we use here `pheatmap`, and show this kind of approach in [heatmap section](https://microbiome.github.io/OMA/microbiome-community.html#composition-heatmap)?

4. 
> The example workflow could also contain
> 
> How to filter rows and cols with very low correlations
> examples on how to visualize some of the most strongly correlated pairs; for checking purposes

I am not quite sure what you meant by this. Should I do cross-correlation analysis between taxa, and filter those taxa that do not have any correlation values over some threshold?

5. This naming system / numbering of files... I'm not sure if current name `40_biclustering.Rmd `is correct

Issue: https://github.com/microbiome/miaViz/issues/39

-Tuomas

### DIFF
--- a/21_microbiome_community.Rmd
+++ b/21_microbiome_community.Rmd
@@ -119,10 +119,10 @@ visualize them with dendrogram.
 
 ```{r pheatmap2}
 # Hierarchical clustering
-taxa_clusters <- hclust(dist(mat), method = "complete")
+taxa_hclust <- hclust(dist(mat), method = "complete")
 
 # Creates a dendrogram
-taxa_dendrogram <- as.dendrogram(taxa_clusters)
+taxa_dendrogram <- as.dendrogram(taxa_hclust)
 
 # Plots it
 plot(taxa_dendrogram)
@@ -132,7 +132,7 @@ Based on dendrogram, we decide to create three clusters.
 
 ```{r pheatmap3}
 # Creates clusters
-taxa_clusters <- cutree(tree = taxa_dendrogram, k = 3)
+taxa_clusters <- cutree(tree = taxa_hclust, k = 3)
 
 # Converts into data frame
 taxa_clusters <- data.frame(clusters = taxa_clusters)
@@ -155,9 +155,36 @@ rowData(tse_phylum_subset)$clusters <- taxa_clusters[order(match(rownames(taxa_c
 rowData(tse_phylum_subset)$clusters
 ```
 
+```{r more_complex_heatmap2}
+# Hierarchical clustering
+sample_hclust <- hclust(dist(t(mat)), method = "complete")
+
+# Creates a dendrogram
+sample_dendrogram <- as.dendrogram(sample_hclust)
+
+```
+
+```{r more_complex_heatmap3}
+# Creates clusters
+sample_clusters <- cutree(tree = sample_hclust, k = 3)
+
+# Converts into data frame
+sample_clusters <- data.frame(clusters = sample_clusters)
+
+# Order data so that it's same as in dendogram
+sample_clusters <- sample_clusters[order(match(rownames(sample_clusters), 
+                                               labels(sample_dendrogram))), , drop = FALSE] 
+
+# Order data based on 
+tse_phylum_subset <- tse_phylum_subset[ , rownames(sample_clusters)]
+
+# Prints taxa and their clusters
+sample_clusters
+```
+
 Now we can create heatmap with additional annotations.
 
-```{r pheatmap5}
+```{r pheatmap4}
 # Creates data frame that includes data on sample type
 sample_types <- unfactor(colData(tse_phylum_subset)$SampleType)
 sample_types <- data.frame(sample_types = sample_types)
@@ -182,10 +209,6 @@ provide functions for more complex heatmaps, such as [_iheatmapr_](https://docs.
 It is also possible to create similar heatmap by just using _ggplot2_.
 
 ```{r more_complex_heatmap1}
-# We already created clusters and dendogram for features
-# taxa_clusters
-# taxa_dendogram
-
 # Add feature names to column as a factor
 taxa_clusters$Feature <- rownames(taxa_clusters)
 taxa_clusters$Feature <- factor(taxa_clusters$Feature, levels = taxa_clusters$Feature)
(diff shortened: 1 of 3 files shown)


antagomir:
Thanks!

**1. cobiclust vs. biclust**

In fact we have two possible use cases for biclustering:
- a) for taxonomic abundance matrix (`cobiclust` is specifically designed for this)
- b) for numeric matrices in general; for instance for correlation matrices of taxa vs. metabolites (`biclust` is suitable for this, with various methods)
- c) more rarely, this could be done for taxa x taxa correlation matrices; then the treatment would as in use case (b)

-> We could list all these use cases, so that it is more clear for the new users what the potential choices could be. Such explanation might help to follow the logic.

Ideally, we could show solutions for both use cases. 

To get started, I suggest to use (b) as the use case, and finalize the example with `biclust`. The `HintikkaXOData()` or `LahtiMLData()` from `microbiomeDataSets` package provide suitable example data sets that have been used for such biclustering in earlier publications.

If `cobiclust` works fine for (a), it would be good to demonstrate that as well. If it does not work well, perhaps we can skip (a) for now, or show a convincing use case with `biclust`.

Remember to cite the original publications, and possibly other biclustering packages that seem relevant (only good ones).


**2. Should biclust be done with clr abundances?**

In use case (a) this is not necessary with `cobiclust` since this is directly applicable to count data. With `biclust` this might help (perhaps with Z transformation on features after CLR on samples) but this is also somewhat problematic because of the pseudocount etc, statistically less sound but perhaps acceptable for exploratory purposes.


**3.   This heatmap includes annotations which are done just by using ggplot2.**

Yes this is good. Either just skip the whole heatmap plotting and refer to the heatmap section for examples; or show a very brief example where the biclusters and indicated on rows and cols. We cannot have a generic solution here.

**4. How to filter rows and cols with very low correlations**

- In use case (a) this means filtering out rare taxa (or, in other words, including prevalent taxa)
- In use cases (b-c) this means filtering out rows and cols that have all correlations lower than a given threshold 

**5. examples on how to visualize some of the most strongly correlated pairs; for checking purposes**

The biclusters cover set of rows, and set of cols; e.g. taxa vs. metabolites. For a given bicluster, we can then estimate sum of abundances, or mean abundance, to represent the overall profile on both margins. Then we could show a scatterplot of total abundances for taxa vs. metabolites for each bicluster.

**6. This naming system / numbering of files... I'm not sure if current name 40_biclustering.Rmd is correct**

This go after `23_multi-assay_analyses.Rmd` i.e. `24_biclustering.Rmd`.

antagomir:
This may contain useful ideas for visualizing bicluster contents in the (possible) use case (a);
`https://julieaubert.github.io/cobiclust/cobiclust-example1.html`

TuomasBorman:
Hi,

now I have taken into account requested changes, Now `cobiclust` works fine, I think that I made some mistake. 

I have included these three cases. Is that scatterplot what you asked?

I added heatmap example to microbiome community section that already had heatmap examples

antagomir:
Ok seems clear, although a bit long. Good to check if it is possible to simplify the writing a bit (for instance also hiding some of the output, or just showing first lines). 

There are many typos, good to check and correct.

Scatterplot: not exactly yet - each bicluster reveals association between rows and columns and in that scatterplot example, this association is for correlation matrix (taxa vs. functional features). This means that every cell in the biclustered matrix corresponds to a correlation between two abundance vectors across samples. In order to understand the bicluster more intuitively, we like to see what kind of associations/correlations there are. We could show correlation scatterplot (taxa vs. functional feature) for every cell of the bicluster, or we can try to summarize the bicluster by just showing mean abundance of row features vs. mean abundance of column features as a scatterplot _across all samples_.  Then the scatterplot should have as many points as there are samples in the original data.

TuomasBorman:
Now I added new scatter plot and removed unnecessary printing

21_microbiome_community.Rmd
```diff
@@ -222,7 +222,31 @@ pheatmap(mat, annotation_row = taxa_clusters,
 ```
 
 In addition to _pheatmap_ package, there are also other packages that
-provide functions for more complex heatmaps, such as [_iheatmapr_](https://docs.ropensci.org/iheatmapr/articles/full_vignettes/iheatmapr.html).
+provide functions for more complex heatmaps, such as [_iheatmapr_](https://docs.ropensci.org/iheatmapr/articles/full_vignettes/iheatmapr.html)
+and 
+[ComplexHeatmap](https://academic.oup.com/bioinformatics/article/32/18/2847/1743594?login=true).
```
antagomir, line 227: Now if **sechm** pkg essentially provides `SE` interface to **ComplexHeatmap**, could we remove the **ComplexHeatmap** examples and just stick to **sechm** for simplicity? Unless there is a particular need to keep both.

With **pheatmap** it might be different, if it provides some other important capabilities that are not available through sechm/ComplexHeatmap?

21_microbiome_community.Rmd
```diff
@@ -222,7 +222,31 @@ pheatmap(mat, annotation_row = taxa_clusters,
...
+                                                        Tongue = "gray")
+
+# Create a plot
+sechm(tse_phylum_subset, 
+      genes = rownames(tse_phylum_subset), 
+      assayName = "clr", 
+      do.scale = TRUE, 
+      top_annotation = c("SampleType"), 
+      cluster_cols = TRUE, cluster_rows = TRUE)
+```
```
antagomir, line 249: perhaps also adding column gaps as this is often needed?

antagomir (approved):

21_microbiome_community.Rmd
TuomasBorman: Currently, there are no examples with `ComplexHeatmap`

I have only used `pheatmap`, `sechm`, and `ggplot`

We could use here only `sechm`, and move other examples to **Visualization chapter** if we decide to create such

antagomir: Ok that's good.

antagomir (approved):

antagomir (approved):

TuomasBorman:
Should we merge this and add new PR for visualization chapter, for instance?

antagomir:
Yes I agree. In fact was just waiting that the automated checks would complete first.

PR #128, opened 2022-01-10
Fix broken link in section on beta diversity

RiboRings:
A link to the section on community typing was broken.


antagomir (changes requested):
In fact, you can cite other chapter with a dynamic link like  `chapter \@ref(example-data)`. Check an example for instance from `04_containers.Rmd`.  The "example-data" tag is defined on another (header) row in that same file but also cross-links between files should be fine.

Can you change like this, then links stay valid even when we change file names and numberings?

RiboRings:
It is clean now. There aren't any other similar cases in the book.

antagomir (approved):

PR #131, opened 2022-01-11
Introduce section on Faith index

RiboRings:
The Faith phylogenetic diversity was previously covered in the [section about diversity](https://github.com/microbiome/OMA/blob/RiboRings/14_alpha_diversity.Rmd) in general (lines 149 - 163). That section is still there, before merging it is a good idea to combine that with the new material I added.


antagomir:
Yes,can you combine?

RiboRings:
It's ready now.

PR #137, opened 2022-01-17
Fix cropped 3D plot

RiboRings:
Hi! I was thinking that, if you also think it's a good idea, I could go through all the other plots in the book and check that they are ok. If they aren't, I can add them to this PR.

For example, I noticed that the last plot in the [section on library size](https://microbiome.github.io/OMA/quality-control.html#library-size) and the plot in the [section on richness](https://microbiome.github.io/OMA/community-diversity.html#richness) encode the same information (the sample type) both as colour and x axis scale. Is that ok or does it cause redundancy?

### DIFF
--- a/12_quality_control.Rmd
+++ b/12_quality_control.Rmd
@@ -288,9 +288,9 @@ plotColData(tse,"sum","X.SampleID", colour_by = "SampleType") +
     labs(y = "Library size (N)", x = "Sample ID") 	    
 ```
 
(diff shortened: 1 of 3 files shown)


antagomir:
yes very good

RiboRings:
Now the colour encodes the barcode information.

antagomir:
Can you make a new fresh PR on this? Something went wrong, perhaps it was outdated. 

Just create new branch from the master, add the new stuff from here and make new PR.

Issue #136, opened 2022-01-17
MOFA example


antagomir:
Add a MOFA workflow example with SE object in chapter "Multi--assay analyses"

PR #138, opened 2022-01-18
MOFA example

ChouaibB:
Hi,

The MOFA workflow examples is added, and the reference to the article.
Issue: #136

### DIFF
--- a/23_multi-assay_analyses.Rmd
+++ b/23_multi-assay_analyses.Rmd
@@ -184,9 +184,9 @@ if(!require(reticulate)){
     install.packages("reticulate")
 }
 
-#reticulate::install_miniconda(force = TRUE)
+reticulate::install_miniconda(force = TRUE)
 reticulate::use_miniconda()
-#reticulate::py_install(packages = c("mofapy2"), pip = TRUE)
+reticulate::py_install(packages = c("mofapy2"), pip = TRUE)
 ```
 
 The `mae` object could be used straight to create the MOFA model:
--- a/23_multi-assay_analyses.Rmd
+++ b/23_multi-assay_analyses.Rmd
@@ -170,9 +170,6 @@ shared among the mutliple (-omics) data sets in hand.
 We use the R [MOFA2](https://biofam.github.io/MOFA2/index.html) package
 for the analysis, and [install](https://biofam.github.io/MOFA2/installation.html) the corresponding dependencies.
 
-**Note:** Comment out the line for installing `miniconda` if it has been installed earlier.
-Similarly for `mofapy2` if it was so.
-
 ```{r, message=FALSE, warning=FALSE}
 if(!require(MOFA2)){
     BiocManager::install("MOFA2")
@@ -189,7 +186,9 @@ reticulate::use_miniconda()
 reticulate::py_install(packages = c("mofapy2"), pip = TRUE)
 ```
 
-The `mae` object could be used straight to create the MOFA model:
+The `mae` object could be used straight to create the MOFA model. Yet, we transform 
+our assays since the model assumes normality per default. Other distributions that
+can be used, include Poisson or Bernoulli.
 
 ```{r, message=FALSE, warning=FALSE}
 library(MOFA2)
@@ -203,17 +202,20 @@ colData(mae)$Diet <- ifelse(colData(mae)$Diet == "High-fat" |
 # which are also in form e.g. "Ambiguous" and "uncultured" taxa
 mae[[1]] <- mae[[1]][!duplicated(rownames(assay(mae[[1]]))), ]
 
-# Transforming microbiome data with clr
-mae[[1]] <- transformCounts(mae[[1]], abund_values = "counts", method = "clr",
-                            pseudocount = 1)
+# Transforming microbiome data with rclr
+mae[[1]] <- transformCounts(mae[[1]], abund_values = "counts", method = "rclr")
 
 # Transforming metabolomic data with log10
 mae[[2]] <- transformSamples(mae[[2]], abund_values = "nmr", method = "log10")
+
+# Transforming biomarker data with z-transform
+mae[[3]] <- transformFeatures(mae[[3]], abund_values = "signals", method = "z", pseudocount = 1)
 
 # Removing assays no longer needed
 assay(mae[[1]], "counts") <- NULL
 assay(mae[[1]], "log10") <- NULL
 assay(mae[[2]], "nmr") <- NULL
+assay(mae[[3]], "signals") <- NULL
 
 # Building our mofa model
 model <- create_mofa_from_MultiAssayExperiment(mae,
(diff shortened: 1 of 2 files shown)


23_multi-assay_analyses.Rmd
```diff
@@ -158,6 +160,120 @@ ggplot(correlation_table, aes(x = Var1, y = Var2, fill = cor)) +
...
+Multi-Omics Factor Analysis [@Argelaguet2018] (MOFA) is
+an unsupervised method for integrating multi-omic data sets in a downstream analysis.
+It could be seen as a generalization of principal component analysis. Yet, with the ability to infer a latent (low-dimensional) representation, 
+shared among the mutliple (-omics) data sets in hand.
+
+We use the R [MOFA2](https://biofam.github.io/MOFA2/index.html) package
+for the analysis, and [install](https://biofam.github.io/MOFA2/installation.html) the corresponding dependencies.
+
+**Note:** Comment out the line for installing `miniconda` if it has been installed earlier.
+Similarly for `mofapy2` if it was so.
```
antagomir, line 174: I think this can be removed.

23_multi-assay_analyses.Rmd
```diff
@@ -158,6 +160,120 @@ ggplot(correlation_table, aes(x = Var1, y = Var2, fill = cor)) +
...
+library(MOFA2)
+# For simplicity, classify all high-fat diets as high-fat, and all the low-fat 
+# diets as low-fat diets
+colData(mae)$Diet <- ifelse(colData(mae)$Diet == "High-fat" | 
+                              colData(mae)$Diet == "High-fat + XOS", 
+                            "High-fat", "Low-fat")
+
+# Removing duplicates at the microbiome data
+# which are also in form e.g. "Ambiguous" and "uncultured" taxa
+mae[[1]] <- mae[[1]][!duplicated(rownames(assay(mae[[1]]))), ]
```
antagomir, line 204: Are these rows also otherwise identical? If yes, can you add an issue to mia, proposing a new function to remove duplicate rows? Like `removeDuplicateRows(tse)`

23_multi-assay_analyses.Rmd
```diff
@@ -158,6 +160,120 @@ ggplot(correlation_table, aes(x = Var1, y = Var2, fill = cor)) +
...
+
+# Transforming microbiome data with clr
+mae[[1]] <- transformCounts(mae[[1]], abund_values = "counts", method = "clr",
+                            pseudocount = 1)
```
antagomir, line 208: Does the result change if we use rclr? That is potentially better than pseudocount+clr but we do not have much experience with it yet.

23_multi-assay_analyses.Rmd
```diff
@@ -158,6 +160,120 @@ ggplot(correlation_table, aes(x = Var1, y = Var2, fill = cor)) +
...
+
+# Transforming metabolomic data with log10
+mae[[2]] <- transformSamples(mae[[2]], abund_values = "nmr", method = "log10")
+
+# Removing assays no longer needed
+assay(mae[[1]], "counts") <- NULL
+assay(mae[[1]], "log10") <- NULL
+assay(mae[[2]], "nmr") <- NULL
```
antagomir, line 216: If this is not really necessary, remove these rows. It complicates the example unnecessarily.

23_multi-assay_analyses.Rmd
```diff
@@ -158,6 +160,120 @@ ggplot(correlation_table, aes(x = Var1, y = Var2, fill = cor)) +
...
+```
+
+Visualizing the variance explained:
+
+```{r, message=FALSE, warning=FALSE, fig.height=8, fig.width=10}
+library(patchwork)
+wrap_plots(
+    plot_variance_explained(model.trained, x="view", y="factor", plot_total = T),
+    nrow = 2
+) + plot_annotation(title = "Varience Explained per factor and assay",
```
antagomir, line 257: typo

antagomir (commented):
Looks good! See comments.

23_multi-assay_analyses.Rmd
ChouaibB: I ran into this problem before, and the reason comes from the `mia::agglomerateByRank` when it is used without `onRankOnly=TRUE` then it produces duplicates. e.g.
image

Earlier in the chapter the agglomeration was done to Genus without the `onRankOnly` argument.
Shall I change the agglomeration done earlier in the chapter, or still open an issue?

23_multi-assay_analyses.Rmd
ChouaibB: I have not tried it yet, I will do it with rclr.

ChouaibB: it did change results considerably (first one with clr, second with rclr):
image
image

23_multi-assay_analyses.Rmd
ChouaibB: It is unfortunately necessary, `MOFA2::create_mofa_from_MultiAssayExperiment` does not give the option to choose which assay from which experiment.

23_multi-assay_analyses.Rmd
antagomir: yes.

23_multi-assay_analyses.Rmd
antagomir: Interesting. I am wondering what MOFA assumes in general about the input data. Would count data be even more appropriate? Does the MOFA have any recommendations?

ChouaibB: The recommendations from MOFA are gaussian likelihoods (the two other options are poisson or bernoulli)

antagomir: are all three data sets approximately Gaussian for the features?

ChouaibB: I unfortunately just by now checked for normality, the two first assays looks approximately normal but not the third:
image

ChouaibB: When I do the features z-transform of the biomarker data I get this:
image
Shall I use that?

antagomir: Yes. Also mention in the text that this is done because the model assumes normality; and just mention those other options that were available.

23_multi-assay_analyses.Rmd
ChouaibB: Should that  "Ambiguous" and "uncultured" taxa be kept?

antagomir: hmmhmm.. for example purposes, you can remove.

antagomir (commented):
Can you confirm when all is ready for merge.

ChouaibB:
yes it is ready, shall I merge it?

PR #145, opened 2022-02-06
Remove DAA example

TuomasBorman:
There is a directory called DAA example. It is created every time `Maaslin2` is used. That directory is not needed to build book so we can delete that.


antagomir:
Ok.

antagomir:
merge after checks pass

PR #157, opened 2022-03-21
Added rda example

TuomasBorman:


20_beta_diversity.Rmd
```diff
@@ -247,6 +247,151 @@ As a final note, `mia` provides functions for the evaluation of additional dissi
...
+# Get assay
+assay <- t(assay(enterotype, "relabundance"))
+# Get colData
+coldata <- colData(enterotype)
+
+# Create a formula
+formula <- as.formula(paste0("assay ~ ", str_c(variable_names, collapse = " + ")) )
+
+# # Perform RDA
+rda <- rda(formula, data = coldata, scale = TRUE, na.action = na.exclude)
```
antagomir, line 285: Have you checked `mia::runRDA` and `mia::calculateRDA`? I wonder if they could do the same directly. May even use the same vegan functionality.

antagomir: Although, if direct calculation can give essential further information, then it might be a useful additional example. Or we could see how to develop runRDA/calculateRDA. The method is so common that it would be useful to automate as much as possible.

20_beta_diversity.Rmd
```diff
@@ -247,6 +247,151 @@ As a final note, `mia` provides functions for the evaluation of additional dissi
...
+
+# Initialize list for p-values
+rda_info <- list()
```
antagomir, line 288: Also wondering if we could/should add or replace example with `dbrda` / `capscale`. These two do the same but one had some added capabilities in terms of interpretation.

>     Distance-based redundancy analysis (dbRDA) is an ordination method
>      similar to Redundancy Analysis (‘rda’), but it allows
>      non-Euclidean dissimilarity indices, such as Manhattan or
>      Bray-Curtis distance. Despite this non-Euclidean feature, the
>      analysis is strictly linear and metric. If called with Euclidean
>      distance, the results are identical to ‘rda’, but dbRDA will be
>      less efficient. Functions ‘capscale’ and ‘dbrda’ are constrained
>      versions of metric scaling, a.k.a. principal coordinates analysis,
>      which are based on the Euclidean distance but can be used, and are
>      more useful, with other dissimilarity measures. The functions can
>      also perform unconstrained principal coordinates analysis,
>      optionally using extended dissimilarities.
>

20_beta_diversity.Rmd
```diff
@@ -247,6 +247,151 @@ As a final note, `mia` provides functions for the evaluation of additional dissi
...
+    new_name <- expr(paste(!!new_name, " (", 
+                           !!format(round( rda_info[variable_name, "proportion"]*100, 1), nsmall = 1), 
+                           "%, ",italic("P"), " = ", 
+                           !!gsub("0\\.","\\.", format(round( rda_info[variable_name, "p_value"], 3), 
+                                                       nsmall = 3)), ")"))
+
+    return(new_name)
+})
+# Add names
+names(vec_lab) <- vec_lab_old
```
antagomir, line 368: Is there a chance to simplify the example? Ideally, the OMA examples would avoid giving too much technical details that are unrelated to the main topic (here RDA). This would be pedagogically better.

antagomir: For figure tuning, one could then cite the visualization chapter, for instance, and more tricks can be there.

20_beta_diversity.Rmd
```diff
@@ -247,6 +247,151 @@ As a final note, `mia` provides functions for the evaluation of additional dissi
...
+           fill = guide_legend("ClinicalStatus"),
+           group = guide_legend("ClinicalStatus"),
+           shape = guide_legend("ClinicalStatus"),
+           x = guide_axis(xlab),
+           y = guide_axis(ylab)) +
+    theme( axis.title = element_text(size = 10) )
+plot
+```
+
+From RDA plto, we can see that only age has significant affect on microbial profile.
```
antagomir, line 393: plto?

antagomir:
`ggord` looks nice.

I also propose to add a new chapter "Constrained ordination methods" after "Community similarity" (which is the current file here).

20_beta_diversity.Rmd
TuomasBorman: Yep, it is possible, I just copy-pasted my code

20_beta_diversity.Rmd
antagomir: I would indeed modify mia functions so that `na.action` can be passed.

If it cannot be done now, can you open an issue there and also refer to the relevant OMA section. Then we can merge this for now (after dealing with the other issues) so that we have something to get started.

20_beta_diversity.Rmd
antagomir: Great. With OMA, also more generally, we really need to pay attention to readability by non-experts.

20_beta_diversity.Rmd
TuomasBorman: It wasn't actually a big thing. Here is PR https://github.com/microbiome/mia/pull/226

Although, I have to still check that it works properly

TuomasBorman: (It casued other error, when different na.action was used, but that it was actrually working correctly)

antagomir: Perfect

20_beta_diversity.Rmd
antagomir: This is rather important, and standard. I suggest to add before the first merge already.

antagomir:
Ah yes you had done this [other dbRDA example](https://microbiome.github.io/OMA/community-similarity.html#testing-differences-in-community-composition-between-sample-groups). Can be also cross-cited. But some of this material, and explanations of adonis2 vs. dbrda vs. capscale could go to the new "Constrained ordination" section.

We should also consider if some of these would deserve their own mia wrapper, a new issue can be opened there is appropriate.

antagomir:
Are all relevant points from the above discussion addressed so that this can be merged?

PR #159, opened 2022-04-21
added biclust to GHA

antagomir:


antagomir:
Aimed to fix the simple failure in the recently merged PR #157

PR #173, opened 2022-06-19
Tasks

antagomir:
I added numbering because with increasing number of tasks I felt that some grouping is necessary.


TuomasBorman:
Yes, good. The intention of mine was actually just remove tasks from table of contents

We can leave them in toc or remove just unnumber after titlrs

Issue #174, opened 2022-06-21
Suggestions


TuomasBorman:
https://microbiome.github.io/OMA/community-similarity.html#other-ordination-methods

rda --> runRDA

TuomasBorman:
maybe simplify the dbrda example??

TuomasBorman:
Remove the group variable?

If one tries to run only on example, the example data set does not have group variable. --> It is not possible to run just one examples at the end of the page

https://microbiome.github.io/OMA/community-similarity.html#explained-variance

TuomasBorman:
https://microbiome.github.io/OMA/community-similarity.html#explained-variance 

There is used vegan functions directly. Change to miaverse functions?

TuomasBorman:
dbRDA does not work when the variable is not binary

https://microbiome.github.io/OMA/community-similarity.html#explained-variance

TuomasBorman:
Add interpretation to examples

for example: https://microbiome.github.io/OMA/community-similarity.html#other-ordination-methods

TuomasBorman:
t-SNE --> UMAP

TuomasBorman:
Remove "FIXME: let us switch to UMAP for the examples?"

TuomasBorman:
To the section that tells about multiassay analyses add a link to mae construction

TuomasBorman:
plotAbundanceDensity: is there an option to add sample annotations (which sample is in which group)

TuomasBorman:
testExperimentCrossCorrelation: option for altExp?

TuomasBorman:
Specify that they are equal with euclidean distances https://microbiome.github.io/OMA/community-similarity.html#testing-differences-in-community-composition-between-sample-groups

TuomasBorman:
Supervised ML?

antagomir:
Yes these are useful suggestions to improve the text.

antagomir:
Can be closed?

TuomasBorman:
These things are taken into account. Can be closed.

PR #175, opened 2022-07-01
Issue 174 - suggestions

TuomasBorman:
https://github.com/microbiome/OMA/issues/174

This is still a draft. I will make modifications based on suggestion described in the issue


TuomasBorman:
> plotAbundanceDensity: is there an option to add sample annotations (which sample is in which group)

I added an example of plotAbundance with sample annotations

PR #183, opened 2022-07-13
add LinDA to DAA; minor updates to material

HenrikEckermann:


30_differential_abundance.Rmd
```diff
@@ -119,11 +119,13 @@ tse <- subsetByPrevalentTaxa(tse, detection = 0, prevalence = 0.1)
...
 ### ALDEx2
 
-In this section, we will show how to perform a simple ALDEx2 analysis. If you wanted to pick a single method, this method could be recommended to use.
+In this section, we will show how to perform a simple ALDEx2 analysis. 
+If you wanted to pick a single method, this method could be recommended to use.
 According to the developers experience, it tends to identify the common
 features identified by other methods. This statement is in line with a recent
-independent evaluation by @nearingMicrobiomeDifferentialAbundance2022.  Please
-also have a look at the more extensive [vignette](https://bioconductor.org/packages/release/bioc/vignettes/ALDEx2/inst/doc/ALDEx2_vignette.html) 
+independent evaluation by @nearingMicrobiomeDifferentialAbundance2022.
```
antagomir, line 126: In general we would recommend using LastnameYear i.e.  Nearing2022 for the reference keys. You could consider changing it here and elsewhere where relevant but no worries if it feels like too much for now.

antagomir (approved):
Looks good to me, thanks for the contribution!

HenrikEckermann:
> Failing checks still. Can you check the reason (we can chat in Slack on the details). We will help when there are problems.

Hi Leo,

yes. I will make sure to fix the issues until it works. Need to install newest version of ANCOMBC on server, then it should run. I also will adapt the citation style. I will let you know in Slack once I am done and merge can be checked again.

PR #187, opened 2022-07-18
HintikkaData

TuomasBorman:
https://github.com/microbiome/mia/pull/282


TuomasBorman:
There are also other datasets used from microbiomeDataSets. We have to think what to do with those. Might be best to use other dataset (some dataset included sequences if I remember correctly. If that is needed in the example of OMA, we might need to include also that dataset in mia)

antagomir:
Yes I agree. The data sets used in OMA and examples are best included in the packages. The microbiomeDataSets could be useful as a general place to share many different ones systematically (also compare to curatedMetagenomicData project, which is dedicated to human only) because package size is very limited. For this reason we will also like to carefully consider which package each data set is best suited.

PR #193, opened 2022-10-02
Add chapter for ML

TuomasBorman:
We do not have chapter for ML. It would be nice to have. 

Additionally, we did not had section for hierarchical clustering which is now added. Now all these clustering text is divided into different chapters 

https://microbiome.github.io/OMA/community-typing.html
https://microbiome.github.io/OMA/biclustering.html
And then we have now also text on hierarchical clustering and k-means

Should we create a new chapter "Clustering" and add hierarchical, k-means and biclustering to there?


antagomir (approved):
Ok! 

We may need to modify/update later, this is a good start.

TuomasBorman:
image

There is a big change. I put old community typing chapter under Community typing (clustering) chapter, because I think it is better to have all these clustering related stuff under one chapter

Is this ok?

antagomir:
Ok for now at least.

antagomir:
(sorry accidentally closed and then reopened)

PR #194, opened 2022-10-19
Refs

antagomir:
New supporting pkgs listed. Good to keep track on these for the time being.


antagomir:
@TuomasBorman this PR seems compatible with master branch however it fails due to ANCOM issues. 

Could we merge regardless of this (does the current master branch pass all tests??). Or perhaps better wait till next Bioc release very soon?

TuomasBorman:
I'm not 100% sure why it fails when master is still passing tests... maybe easiest is to wait the release

antagomir:
Might work now.

PR #204, opened 2022-12-14
update rda example

TuomasBorman:


20_beta_diversity.Rmd
```diff
@@ -273,50 +273,66 @@ variable_names <- c("ClinicalStatus", "Gender", "Age")
 # Apply relative transform
 enterotype <- transformSamples(enterotype, method = "relabundance")
 
-# Get assay
-assay <- t(assay(enterotype, "relabundance"))
-# Get colData
-coldata <- colData(enterotype)
-
 # Create a formula
```
antagomir, line 276: Just a bit more explanation to support naive users might be useful here; like assay is the response variable, and the explanatory variables would be listed in variable_names but do not have to be named here yet.

20_beta_diversity.Rmd
```diff
@@ -273,50 +273,66 @@ variable_names <- c("ClinicalStatus", "Gender", "Age")
...
-      paste0("P-value (PERMANOVA ", permutations, " permutations)") )
-# Print info
-kable(rda_info_clean)
+rda <- calculateRDA(enterotype, assay_name = "relabundance",
+                    formula = formula, distance = "bray", na.action = na.exclude)
+# Get the rda object
+rda <- attr(rda, "rda")
+# Calculate p-value and variance for whole model
+set.seed(436)
+permanova <- anova.cca(rda, permutations = 999)
```
antagomir, line 286: I would show example with 99 permutations, and mention in comments that it is usually recommended to have more in practice, like 999.

antagomir: This just to keep the already notable computing load of the book somehow limited. This general comment applies to all examples.

20_beta_diversity.Rmd
```diff
@@ -273,50 +273,66 @@ variable_names <- c("ClinicalStatus", "Gender", "Age")
...
+rda_info <- rbind(rda_info, permanova)
+
+# Add info about total variance
+rda_info[ , "Total variance"] <- rda_info["Model", 2] +
+    rda_info["Residual", 2]
+
+# Add info about explained variance
+rda_info[ , "Explained variance"] <- rda_info[ , 2] / 
+    rda_info[ , "Total variance"]
+
```
antagomir, line 304: Just wondering if we could standarize these steps even better and incorporate them into the function `calculateRDA`? While maintaining some standarization among the similarly named runRDA, runMDS, runPCA etc. functions (or creating another wrapper with a different name). The point being, if this is an essential and common task that comes up all the time in taxonomic analyses, why not streamline it.

antagomir: In some sense I symphatize the idea that users calculate everything themselves. 

But for naive users that might be too much, and for experienced users that could cause unnecessary work load as they would just like to not loose time and avoid possibilities for manual errors. And documentation could also show the more detailed ways or link to material that shows.

20_beta_diversity.Rmd
```diff
@@ -273,50 +273,66 @@ variable_names <- c("ClinicalStatus", "Gender", "Age")
...
+        set.seed(413)
+        temp <- anova(
+            betadisper( 
+                vegdist(assay, method = "bray"),
+                group = coldata[[variable_name]] ),
+            permutations = permutations )["Groups", "Pr(>F)"]
+    }
+    # Add info to the list
+    homogeneity[[variable_name]] <- temp
+}
```
antagomir, line 330: Same, would it be wise to consider standardization.

antagomir (approved):
Approved but comments to consider. If not solved now, perhaps open issues.

20_beta_diversity.Rmd
TuomasBorman: Yes, there are advantages and I think this could be standardized

PR #206, opened 2023-01-10
var expl. + p-val from --> anova.cca

ChouaibB:
At the example of RDA, the calculation of variance explained for whole model and per each variable plus their corresponding p-values, is switched to using `anova.cca(rda, by="margin")`.

### DIFF
--- a/20_beta_diversity.Rmd
+++ b/20_beta_diversity.Rmd
@@ -285,40 +285,29 @@ coldata <- colData(enterotype)
 # Create a formula
 formula <- as.formula(paste0("assay ~ ", str_c(variable_names, collapse = " + ")) )
 
-# # Perform RDA
+# Perform RDA
 rda <- rda(formula, data = coldata, scale = TRUE, na.action = na.exclude)
 
-# Initialize list for p-values
-rda_info <- list()
-# Name for storing the result
-variable_name <- "all"
-# Calculate and store p-value, and other information
-rda_info[[variable_name]] <- c(constrained = rda$CCA$tot.chi, 
-                               unconstrainded = rda$CA$tot.chi, 
-                               proportion = rda$CCA$tot.chi/rda$CA$tot.chi, 
-                               p_value = anova.cca(rda)["Model", "Pr(>F)"] )
-
-# Loop through variables
+# retrieving variances explained and p-values for the whole model and for each variable
 permutations <- 999
-for( variable_name in variable_names ){
-    # Create a formula
-    formula <- as.formula(paste0("assay ~ ", variable_name) )
-    # Perform RDA
-    rda_temp <- rda(formula, data = coldata, scale = TRUE, na.action = na.exclude)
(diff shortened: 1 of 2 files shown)


ChouaibB:
I just realized this is a duplicate of #204 .
Otherwise, instructing mofa installation at [Multi-assay analyses](https://github.com/microbiome/OMA/blob/master/23_multi-assay_analyses.Rmd#L160) to use python 3.6 could solve the workflow run problem.

antagomir:
Great, we should close the current PRs asap..!

Sorry for the hassle, let us first merge #204 and then see if this has some additional useful contents.

COuld you make PR on the Python suggestion?

ChouaibB:
> Great, we should close the current PRs asap..!
> 
> Sorry for the hassle, let us first merge #204 and then see if this has some additional useful contents.
> 
> COuld you make PR on the Python suggestion?

Sure.

TuomasBorman (approved):
https://github.com/microbiome/OMA/pull/204 merged

Sorry that it took so long; chnages in this PR looks good; sync with master and check that everything stays correct

ChouaibB:
> #204 merged
> 
> Sorry that it took so long; chnages in this PR looks good; sync with master and check that everything stays correct

Hi, sorry got confused for a moment, which changes would you like to keep and/or discard? The table with variance and p-vals?

antagomir:
The earlier #204 is now merged. It would be good to check if this #206 has additional useful content to keep, to complement, clarify, or streamline the example with useful material. If uncertain, we can discuss here about the choices, or sync and update the PR and we can discuss in the code review.

ChouaibB:
Yes please, leave comments.
This one is rda but the merged one is dbRDA, variances are calculated differently, and the merged one has additional info about homogeneity.

TuomasBorman:
Now it seems that the latest commit removed the whole file? Can you revert the commit and merge the master to your branch

ChouaibB:
> Now it seems that the latest commit removed the whole file? Can you revert the commit and merge the master to your branch

I have reverted the file to the previous rda example there was initially. I didn't merge the master, as the example existing in master now is dbRDA and explained variances are calculated differently (ofc creating conflict if to be merged); and different results...
If you would like to keep the dbRDA existing example, there is no point of this PR.
I think existing example is good, enabling distance metrics in the analysis, additionally how the homogeneity calculation is done. 
However, about the variance explained calculations, I was not sure the right way of doing it, but rather based on the [comment at the source code of anova.cca](https://github.com/vegandevs/vegan/blob/master/R/anova.ccabyterm.R#L41). I am still looking for more resources explaining the procedure.

antagomir:
If we can use anything from this PR to streamline or otherwise improve the current OMA example then it would be worthwhile. If you can check and conclude that would be great. Otherwise, I should have a look but this may take some days at least.

antagomir:
Let us close this one. We can return to refine later as necessary.

PR #207, opened 2023-01-11
workflow fix; mofa --> python 3.6

ChouaibB:
Fixing workflow fails by instructing mofa installation at [Multi-assay analyses](https://github.com/microbiome/OMA/blob/master/23_multi-assay_analyses.Rmd#L160) to use python 3.6.
(related to discussion at #206 )

### DIFF
--- a/23_multi-assay_analyses.Rmd
+++ b/23_multi-assay_analyses.Rmd
@@ -157,7 +157,7 @@ if(!require(reticulate)){
 
 reticulate::install_miniconda(force = TRUE)
 reticulate::use_miniconda(condaenv = "env1", required = FALSE)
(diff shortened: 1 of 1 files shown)


antagomir (approved):
Seems good. Newer versions of Python (3.11 etc) do not work? If this solves our case it is good but it is also risky if we have to enforce old versions. Ok for now if the newer ones do not work.

ChouaibB:
> Seems good. Newer versions of Python (3.11 etc) do not work? If this solves our case it is good but it is also risky if we have to enforce old versions. Ok for now if the newer ones do not work.

Yes they don't. Newer versions, installed newer packages with new/modified methods upon which mofa2 is not built. Perhaps, newer releases of mofa2 would need to revisit this example once again.

PR #208, opened 2023-01-27
Fix broken cross-references

RiboRings:
Hi! I fixed the last couple of cross-references that weren't working.


antagomir:
Thanks!

PR #214, opened 2023-02-21
Delete duplicate lines

RiboRings:
Hi!

I noticed that in [PERMANOVA comparison](https://microbiome.github.io/OMA/extras.html#permanova-comparison) section there are some duplicate lines, so I deleted them.

Just in case if you could double check.
Cheers!

### DIFF
--- a/97_extra_materials.Rmd
+++ b/97_extra_materials.Rmd
@@ -110,7 +110,7 @@ vars <- c("Nationality", "Gender", "ClinicalStatus")
 by_args <- factor(c("terms", "margin"))
 
 var_perm <- permutations(n = 3, r = 3, vars)
-formulas <- apply(var_perm, 1, function(row) reduce(row, function(x, y) paste(x, "+", y)))
+formulas <- apply(var_perm, 1, function(row) purrr::reduce(row, function(x, y) paste(x, "+", y)))
 
 terms_df <- data.frame("Formula" = formulas,
                        "ClinicalStatus" = rep(0, 6),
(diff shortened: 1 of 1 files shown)


antagomir (commented):
The covariates seem to be in different order in the "duplicate" examples, so the duplicates are not identical. Does the order influence p-values in this case? If yes, then at least a note should be added mentioning that the order of covariates affects the outcome.

RiboRings:
Could you give me a hint how to use `as.function`?

I tried the following inside a for-loop, where my_grid$Var1 are all the combinations of explanatory variables, and my_grid$Var2 are all the `by` arguments.

```
 my_ad <- adonis2(formula = as.function(paste0('t(assay(enterotype, "relabundance")) ~ ', my_grid$Var1[idx])), 
                   by = as.character(my_grid$Var2[idx]),
                   data = colData(enterotype),
                   permutations = 99)
           
```

I get the error: `Error in as.function.default(paste0("t(assay(enterotype, \"relabundance\")) ~ ",  : 
  list argument expected`

RiboRings:
Never mind, it seems to work with `as.formula`

antagomir (approved):

antagomir (approved):

antagomir:
Unnamed chunk causes some error. 

Could you add names for the chunks, and see if the error could be spotted..?

RiboRings:
Yes, thanks for the clue!

antagomir:
Awesome!

PR #215, opened 2023-02-23
fixed broken code + reference to new feature of ancombc2 function

HenrikEckermann:
I promised a while ago I will go over the DAA section in order to reference to new developments for ANCOMBC. ANCOMBC has become quite complex by now with lots of functionality. The simple example that we cover in the OMA book does not need this complexity. So, I kept the simple example but make the reader aware of possibilities/new developments and reference to the elaborate tutorials by the developer. I think that is the best solution for now. 

Regarding the other changes:
I noticed the code was broken due to package updates. Several packages handled the input differently compared to months ago. I assume that won't be the last time this happens. For now the code should be fixed but I left eval = FALSE as I am not sure when you want to make it evaluate on the server again.

### DIFF
--- a/30_differential_abundance.Rmd
+++ b/30_differential_abundance.Rmd
@@ -79,15 +79,11 @@ library(Maaslin2)
 library(MicrobiomeStat)
 library(knitr)
 library(tidyverse)
+library(ANCOMBC)
 
 # set random seed because some tools can randomly vary and then produce 
 # different results:
 set.seed(13253)
-
-# For ANCOMBC we need development version
-# Can be installed with:
-# remotes::install_github("FrederickHuangLin/ANCOMBC")
-library(ANCOMBC)
 
 # we use a demo dataset and restrict it to two geo locations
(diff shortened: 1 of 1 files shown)


antagomir:
Very useful, thanks.

The evals were set to FALSE last week when I needed to get some other updates live urgently for a workshop. If you could release the examples for this PR with `eval=TRUE`  we could test if they pass the GHA checks.

On the same go:

1) If some DA examples are very slow to run we should reconsider how to include in OMA. Not sure if this is the case but if you notice such behavior we could discuss.

2) Methods may have differences in whether they can treat only discrete or also continuous variables; and whether they can include covariates. I was starting to wonder if it would be useful to include here a summary table for these (and possibly other) differences to facilitate the choice of a suitable method. Might be easy to add but harder to maintain. Although the maintenance burden applies to all material. Your comments welcome.

antagomir:
Ok I merge this now anyway, we can continue the discussion later.

Issue #238, opened 2023-04-05
Simplify library load


antagomir:
We could simplify 

```
if( !require(vegan) ){
    BiocManager::install("vegan")
    library("vegan")
}
```

with just 

```
library("vegan")
```

or 

```
vegan::vegdist
```

etc.

This would improve readability, simplify examples. If it is necessary, one could add a section or cite some basic R resources that show how one can install packages. In general, we could expect that the users know how to do that when necessary.

antagomir:
Ready? (close if yes)

PR #270, opened 2023-05-29
Update reproducible reporting and outreach

RiboRings:
Hi! I moved the reproducible reporting section from course webpage to OMA. Something was already in OMA, so I just modified it. I also added outreach section in resources chapter with quarto presentations.

### DIFF
--- a/04_containers.Rmd
+++ b/04_containers.Rmd
@@ -271,6 +271,13 @@ help(GlobalPatterns)
 ```
 
 
+#### HintikkaXOData {#hintikka-desc}
+
+[HintikkaXOData](https://microbiome.github.io/microbiomeDataSets/reference/HintikkaXOData.html)
+is a MAE derived from a study about the effects of fat diet and prebiotics on the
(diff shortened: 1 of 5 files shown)


antagomir:
Seems very good.

antagomir (commented):
Can you check issue #238 and update accordingly.

Also possible to update the other parts in the same way for the same PR

RiboRings:
> Can you check issue #238 and update accordingly.
> 
> Also possible to update the other parts in the same way for the same PR

The problem is that without
`if(!require(GUniFrac)){
  install.packages("GUniFrac")
}`

an import error arises. Or am I missing something?

antagomir:
If that really is the case then no problem to have it.

I just wonder how common that is.

My default solution would be to add this in [.github/workflows](https://github.com/microbiome/OMA/blob/master/.github/workflows/build-book-release.yml) that describes the OMA conversion. I can see that GUnifrac is already there so I am not sure what the problem is. 

Ok we can merge in this case as this passes otherwise. We can continue to troubleshoot issue #238 separately. 

Thanks!

PR #279, opened 2023-06-08
Expand exercises on and polish multi-omics chapter

RiboRings:


98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
-2. Combine TreeSE into MAE.
-3. Check that each individual experiment of MAE equals corresponding TreeSE.
-4. Take a subset of MAE (e.g., 10 first samples), and observe the subsetted MAE.
-
-Useful functions: DataFrame, TreeSummarizedExperiment, matrix, rownames, colnames, MultiAssayExperiment, ExperimentList, SimpleList
-
+The following exercise walks you through disassembling a MAE object in
+order to store it as multiple separate csv files.
+
+1. Import the microbiomeDataSets package
```
antagomir, line 297: Note: 

HintikkaXOData is a special case that is available both in the mia package as well as through microbiomeDataSets. We can do either - - 

without download over the web:

```
data("HintikkaXOData", package="mia")
```

or 


with download over the web:

```
mae <- microbiomeDataSets::HintikkaXOData()
```

I would prefer the first approach since relying on network is not always good.

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+2. Store the HintikkaXOData MAE into a variable named `mae`
+3. Which and how many experiments are contained by the MAE?
+4. Split the MAE into TreeSEs named `microbiota`, `metabolites` and `biomarkers`
```
antagomir, line 300: Or: "Extract the individual experiments from the MAE data container into distinct data objects."

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+5. Which and how many assays are contained by each TreeSE?
```
antagomir, line 301: Enough to ask just for one experiment?

Experiments could in principle have other format than TreeSE. Therefore I would call them "experiments" rather than TreeSE's.

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+6. Store the assays of each TreeSE into a variable named `<TreeSE name>_assays`
```
antagomir, line 302: I would remove this, or or move it into question that asks to extract for instance abundance table out of the microbiome expriment.

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+7. Write a csv file for each assays object with the `write.csv` function
```
antagomir, line 303: doing this for one assay could be enough for exercise. They can do the rest if they want to but I would not make it a too repetitive example.

98_exercises.Rmd
antagomir: Pedagogically thinking I think it might be easier to teach them first separately how to create TreeSE object from scractch. Then, MAE construction could be based on just 2-3 ready made TreeSE objects.

How about adding example on how to do this kind of thing with a single TreeSE (in some earlier exercise chapter)? And then here just suggesting to do that exercise for the individual experiments of the MAE, and then in this example focusing exclusively on how MAE is constructed from ready-made TreeSEs? 

I like the idea but wondering if this is getting too broad task at once?

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+8. Besides experiments, MAEs also include a sampleData and a sampleMap, which
+   are accessible with `colData(mae)` and `sampleMap(mae)`, respectively.
+   Save also each of these two elements into a csv file.
+
+Well done! You just splitted a MAE into its components and stored them as csv files.
+[This script](https://github.com/JuliaTurkuDataScience/MicrobiomeAnalysis.jl/blob/main/src/assets/XO_preprocess.R)
```
antagomir, line 309: This refers to Julia code. Is this good for R tutorial?

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+3. Read in the sampleData and the sampleMap and store them into the
+   variables `sample_data` and `sample_map`, respectively
+4. Combine the TreeSEs, the sampleData and the sampleMap into a MAE named `mae`
+   with the `MultiAssayExperiment` function, where the first argument is an
+   `ExperimentList` of your TreeSEs, the second is colData and the third is
+   sampleMap
+5. Make sure that the MAE experiments are identical to the original TreeSEs.
+
+Good job! Now you are aware of how MAEs are built and we can proceed to some
+explorative exercises.
```
antagomir, line 328: Nice

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
 
-1. Load experimental dataset from microbiomeDataSets (e.g., HintikkaXOData).
-2. Analyze correlations between experiments. (Taxa vs lipids, Taxa vs biomarkers, Lipids vs biomarkers)
-3. Agglomerate taxa data.
-4. Apply CLR to taxa data, apply log10 to lipids and biomarkers.
-5. Perform cross-correlation analyses and visualize results with heatmaps. (Use Spearman coefficients)
-6. Is there significant correlations? Interpret your results.
-
-Useful functions: pheatmap, ComplexHeatMap::Heatmap, ggplot, transformCounts, testExperimentCrossAssociation
+Here we learn some explorative methods and apply to the HintikkaXOData. If you
```
antagomir, line 332: i think it should be "exploratory"

antagomir: No "the" before names (like HintikkaXOData)

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+solved the previous exercise, you can use the MAE you created and skip the first
+step.
```
antagomir, line 334: Or you can just use data(HintikkaXOData, package="mia")

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+
```
antagomir, line 335: I would have the initial CSV saving and loading exercises _after_ the data manipulation exercises. This way students can first get motivation on why this thing can be useful, and then they may have more motivation to learn details on how to create it on your own.

98_exercises.Rmd
antagomir: In fact I am not sure if we need to have exercise on how to store CSV files here in this context. It is basic R skill, not MAE specific skill. We could also just give web location for the CSV files and suggest to try TreeSE -> MAE construction based on those directly

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+1. Store the HintikkaXOData MAE into a variable named `mae`
+2. Which experiments make up the MAE? How many samples and features are contained
+   by each experiment? You can check that with `dim`, `nrow` and `ncol`
```
antagomir, line 338: I would start with these exercises for MAE

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+3. Agglomerate the microbiota experiment by Genus and store it into a new TreeSE
+   named `microbiota_genus`. How many features remain after agglomerating?
+4. Add `microbiota_genus` as a new experiment to `mae`. Note that new experiments
```
antagomir, line 341: These are more advanced, I would split this into its own exercise.

antagomir: In fact this is not good example because we are suggesting to store different levels of taxonomic tables into altExp, see Table 3.1 https://microbiome.github.io/OMA/containers.html#data-science-framework

antagomir: You could ask to add the genus level table into a new altExp within the microbiome experiment in mae..?

antagomir: For adding new MAE components, how about just construcing one MAE with two components first (microbiota + metabolites?) and then adding third in a separate step?

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+   can be added to a MAE through simple concatenation with `c(mae, ...)`
+   
+So far so good. You explored the MAE, agglomerated one of its experiments and
+added it to the container. Now we will perform a cross-correlation analysis
+between two of the experiments in the MAE.
```
antagomir, line 346: Could we have titles for these exercises? Can be subsubsubtitle if necessary. It will make it easier to refer to exercises by number + name

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+   which of their assays with `assay.type1` and `assay.type2`
+3. What does the output look like? By default, correlation is measured in terms
+   of Kendall tau coefficients. Repeat point 2, but this time change the method
+   to Spearman coefficients.
+4. Are you able to infer significance from the output? In order to also obtain
+   p-values from the cross-correlation analysis, repeat point 2 with the
+   additional argument `test_significance = TRUE`
+5. Visualize results with a heatmap similarly to the example in section
+   \@ref(cross-correlation). Do you see any significant correlations?
+   Interpret your results.
```
antagomir, line 365: Let us have separate exercises on adding transformations or other manipulations in MAE.

The cross correlation experiment could be simplified by not including new transformations. 

Let us keep each exercise more focused on learning one thing well. We can have multiple exercises that partially complement each other.

antagomir: -> so this exercise could be splitted a bit.

98_exercises.Rmd
```diff
@@ -291,24 +291,87 @@ See the OMA book for examples.
...
+
+Congratulations! You performed a cross-correlation analysis on a MAE data
+container. Feel free to further test your skills with the following tips:
+
+* Perform cross-correlation analysis between the remaining experiments
+  (microbiota vs metabolites and metabolites vs biomarkers)
+* Focus on the significant correlations between taxa and biomarkers and
+  predict those biomarkers from taxa by means of linear regression
+* Move further onto multi-omics factor analysis (\@ref(mofa)).
```
antagomir, line 374: Add numbering for examples, and tips on suitable functions.

antagomir (changes requested):

antagomir (changes requested):
Well done - now check the comments for some polishing.

antagomir (changes requested):
Well done - now check the comments for some polishing.

98_exercises.Rmd
RiboRings: [XO_preprocess.R](https://github.com/JuliaTurkuDataScience/MicrobiomeAnalysis.jl/blob/main/src/assets/XO_preprocess.R) is an R script and is basically the solution to the exercise.

But if you think it is confusing because it is in a Julia package, then I'll remove it (I don't think anyone would notice though).

antagomir: I think we should not mix R and Julia material this way. 

Or what would be the argument for keeping that R script in Julia package, instead of OMA repository or mia R package repository?

RiboRings: Because I used it in the documentation to show how to disassemble MAEs in R and import them in Julia: https://juliaturkudatascience.github.io/MicrobiomeAnalysis.jl/dev/example3/#Importing-MAE

RiboRings: Would there be a more suitable place in OMA or mia?

98_exercises.Rmd
```diff
@@ -435,11 +435,14 @@ customise the correlation method and significance testing used for the analysis.
 
 ### Multi-omic Factor Analysis (MOFA)
 
-We further move onto an exercise on MOFA.
+TO DO
```
antagomir, line 438: I would remove "TO DO"

98_exercises.Rmd
```diff
@@ -435,11 +435,14 @@ customise the correlation method and significance testing used for the analysis.
...
 
-* Move further onto multi-omics factor analysis (\@ref(mofa)).
+We further move onto an exercise on MOFA, which is explained in chapter \@ref(mofa).
```
antagomir, line 440: Does this mean that MOFA exercise is in the MOFA chapter and not in the exercise chapter?

Wouldn't it be beneficial to have all exercises in one place (exercise chapter), and then refer to those in the text? Or alternatively have all exercises embedded in the text chapters (which I think will be more difficult to maintain and develop further).

98_exercises.Rmd
```diff
@@ -435,11 +435,14 @@ customise the correlation method and significance testing used for the analysis.
...
 
-Congratulations! You conducted a MOFA bla bla bla.
+1. Import the mia package, load HintikkaXOData with `data` and store it into a
+   variable named `mae`
+
+Congratulations! You conducted a MOFA.
```
antagomir, line 445: How 445 is related to 442-443?

98_exercises.Rmd
RiboRings: No sorry. It just means that I still have to make an exercise about it (if necessary). Otherwise I'll just delete this part

antagomir: This is ok.

Issue #278, opened 2023-06-07
References


antagomir:
In OMA, section 16, and perhaps also other sections, contain small URL issue, the https:: starting part is duplicated.
"https://doi.org/https://doi.org/10.12688/f1000research.26669.2."

-> Check and fix from bib file.
-> Primarily use DOI as the URL for every publication, where possible

Issue #276, opened 2023-06-07
Add MAE info in chapter 16.1.1


antagomir:
Add MultiAssayExperiment (MAE) info in [chapter 16.1.1](https://microbiome.github.io/OMA/resources.html#data-containers-1).

Also change the section title from "Resources for TreeSummarizedExperiment" into  "Data container documentation"

Issue #273, opened 2023-06-05
Acknowledgments


antagomir:
Similar approach [with OSCA](https://bioconductor.org/books/release/OSCA/contributors.html) could work providing acknowledgments for contributors.

PR #281, opened 2023-06-09
 #276,  #278, #273, #290

ake123:
- Changed the section title from `"Resources for TreeSummarizedExperiment"` into `"Data container documentation"`
- Added MultiAssayExperiment info in chapter 16.1.1
- Edited the `book.bib` section to add `MultiAssayExperiment`
- References #278 (fix from bib file)
- Acknowledgments #273
- Broken links #290


book.bib
```diff
@@ -548,6 +548,21 @@ @Manual{R-TreeSummarizedExperiment
   note =	 {R package version 1.6.2},
 }
 
+@Manual{R-MultiAssayExperiment,
```
antagomir, line 551: We should not use "-" in bibtex keys. It may cause problems sometimes. 

-> Replace "-" in bibtext keys with "_" or "".

antagomir: (in any reference)

book.bib
```diff
@@ -548,6 +548,21 @@ @Manual{R-TreeSummarizedExperiment
...
+  title =	 {MultiAssayExperiment: Software For The Integration Of Multi-Omics
```
antagomir, line 552: Put MultiAssayExperiment in curly brackets "{MultiAssayExperiment}" -> this will preserve capitalization in different output layouts. That is essential because this is a name.

book.bib
```diff
@@ -548,6 +548,21 @@ @Manual{R-TreeSummarizedExperiment
...
+                  and Benjamin Haibe-Kains and Kasper Hansen and
+                  Hanish Kodali and Marie Stephie Louis and Arvind
+                  Singh Mer and Markus Reister and Martin Morgan and
+                  Vincent Carey and Levi Waldron},
+  year =	 2017,
+  note =	 {R package version 1.26.0},
+}
+
 @Manual{R-XVector,
   title =	 {XVector: Foundation of external vector
```
antagomir, line 567: same

antagomir (commented):
Very useful. Just some relatively straightfwd additional updates.

book.bib
```diff
@@ -549,7 +549,7 @@ @Manual{R-TreeSummarizedExperiment
 }
 
 @Manual{R-MultiAssayExperiment,
-  title =	 {MultiAssayExperiment: Software For The Integration Of Multi-Omics
+  title =	 {{MultiAssayExperiment}: Software For The Integration Of Multi Omics
```
antagomir, line 552: The original publication seems to use word "multiomics" https://pubmed.ncbi.nlm.nih.gov/29092936/ 

-> fix

book.bib
ake123: I see lot's of   "-" in the bibtext within the document. Should all be removed?

antagomir: It only matters when it is in the key, meaning 

@article{name-with-dash

@manual{name-with-dash

etc.

In article titles etc. it is ok.

Yes, for all keys it would be good to change, and probably easiest to do for all at once.

On the other hand, if we do not have problems with it now in current setup I think we can leave this later in the end. Let's keep in mind anyway.

antagomir:
Conflicts to resolve-

PR #285, opened 2023-06-16
Updated intro to bluster to include the wrapper

BananaCancer:
I changed the code and comments to mention the wrapper and how to use it. I haven't mentioned the altExps so far because I don't think that's the main use, but tell me if I should.
Examples are given for the samples but it's explained for the features. The storing of the metadata is also explained as well as the naming of cols, rows and metadata.
Don't hesitate to comment.

### DIFF
--- a/24_clustering.Rmd
+++ b/24_clustering.Rmd
@@ -255,7 +255,7 @@ Show Laplace approximation (model evidence) for each model of the k models.
 
 ```{r}
 library(miaViz)
-plotDMNFit(tse_dmn, type = "laplace")
+plotDMNFit(tse_dmm, type = "laplace")
 ```
 
 Get the model that has the best fit.
--- a/24_clustering.Rmd
+++ b/24_clustering.Rmd
@@ -381,6 +381,7 @@ ratio (rclr) assay data. Then plotting the communities using UMAP
 [@McInnes2018] ordination as a visual exploration aid. Let us cluster the 
 `enterotype` dataset.
 
+
 ```{r, message=FALSE, warning=FALSE}
 library(bluster)
 library(patchwork) # For arranging several plots as a grid
@@ -411,7 +412,7 @@ plots <- lapply(k, ClustAndPlot)
 plots[[1]] + plots[[2]] + plots[[3]] + plots[[4]] + plot_layout(ncol = 4)
 ```
 
-Similarly, the _`bluster`_ [@R-bluster] package offers clustering
+Similarly, the _`bluster`_ [@R_bluster] package offers clustering
 diagnostics that can be used for judging the clustering quality (see
 [Assorted clustering
 diagnostics](http://bioconductor.org/packages/release/bioc/vignettes/bluster/inst/doc/diagnostics.html)).
--- a/24_clustering.Rmd
+++ b/24_clustering.Rmd
@@ -26,10 +26,7 @@ to do when using this package is to load it, and transform the data if
 necessary, depending on your analysis goals.
 
 ```{r load_bluster}
-# Install dependencies (to remove in the next bioc release)
-devtools::install_github("LTLA/bluster") 
-
-# Load dependencies
+# Load dependencies (remove github dependency in the next bioc release)
 library(bluster)
 
 # Apply transformation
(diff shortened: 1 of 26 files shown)


24_clustering.Rmd
```diff
@@ -18,7 +18,7 @@ features/samples that share pattern.  For example, with clustering, we
 can find group of samples that share similar community
 composition. There are multiple clustering algorithms available.
 
-## Introduction to bluster package
+## Introduction to bluster package and wrapper
```
antagomir, line 21: I wouldn't mention "wrappers" in this text. The typical user does not have to care what kind of function it is, that would be more like developer stuff.

antagomir, line 21: Another stylistic suggestion - I would generally avoid mentioning individual packages in titles. Instead I would focus on the conceptual idea. Here, an alternative title could be like: "Custom tools"

24_clustering.Rmd
```diff
@@ -18,7 +18,7 @@ features/samples that share pattern.  For example, with clustering, we
...
 
 *bluster* is a Bioconductor package specialized in clustering. It offers
```
antagomir, line 23: The *bluster* Bioconductor package provides specialized tools for clustering data in the `SummarizedExperiment` container.

24_clustering.Rmd
```diff
@@ -33,7 +33,10 @@ library(bluster)
 tse <- transformCounts(tse, method = "relabundance")
 ```
 
-The clustering can be done on features or samples.
+*bluster*'s main function `clusterRows`, used to do the clustering can be 
+used via mia's `cluster` function.
```
antagomir, line 37: Reconsider how much to explain function names in the text. They will come in the code examples in any case. There are no strict rules on this but the main text should be fluent and easy to read, and many function names make it more technical and we want to go beyond a technical manual in this online book.

antagomir:
General request to add to this PR: the current chapter 10.5 in OMA "10.5 Community Detection" 

- reconsider the title: does "Community detection" here somehow differ from the other clustering methods? Are they not all for community detection? Should we instead name this by the clustering method, as in the other subchapters? 

- the enterotype data set is originally from the phyloseq package I think, and based on another publication (Arumugam et al. 2011). The mia package provides all relevant references, so change the text "In the following demonstration we use the enterotype dataset from the (Ernst, Shetty, and Lahti 2020) package." into "Let us cluster the enterotype dataset."

- the figures with different values of k could be put on the same row for brevity

BananaCancer:
> General request to add to this PR: the current chapter 10.5 in OMA "10.5 Community Detection"
> 
>     * reconsider the title: does "Community detection" here somehow differ from the other clustering methods? Are they not all for community detection? Should we instead name this by the clustering method, as in the other subchapters?
> 
>     * the enterotype data set is originally from the phyloseq package I think, and based on another publication (Arumugam et al. 2011). The mia package provides all relevant references, so change the text "In the following demonstration we use the enterotype dataset from the (Ernst, Shetty, and Lahti 2020) package." into "Let us cluster the enterotype dataset."
> 
>     * the figures with different values of k could be put on the same row for brevity

I'll have a look at that, and I also want to review the rest of the clustering chapter as it can be improved with the upcoming changes to bluster and mia. That being said, I need the changes to be merged in order to use them in the examples.

BananaCancer:
Is this version good?
I'll check it again once the DMM-param in bluster and the cluster wrapper in mia have been merged.

antagomir (approved):
Seems good to me.

24_clustering.Rmd
```diff
@@ -402,30 +370,23 @@ euclidean_dmm_plot <- ggplot(
...
-among which graph-based are present, enabling the community detection
-task.
-
-Installing package:
-
-```{r}
-library(bluster)
-```
+Here, we will be using the `cluster` function with a graph-based clustering 
+function, enabling the community detection task.
```
antagomir, line 381: Is there some recent microbiome literature showing that this method has some advantages compared to the other clustering methods? Otherwise we could potentially remove this example case to keep contents focused.

24_clustering.Rmd
```diff
@@ -78,43 +74,15 @@ plotReducedDim(tse, "MDS", colour_by = "clusters")
...
-```{r scran_sample_clust}
-library(scran)
-pam.out <- clusterCells(tse,
-  assay.type = "relabundance",
-  BLUSPARAM = PamParam(centers = 5)
-)
-```
+Check dedicated part to see other examples.
 
 ### Taxa clustering
```
antagomir, line 79: Remove "Taxa clustering" section. The chapter 6.3 discusses those things. You can briefly refer to that section possibly, but that is not critical. Let us avoid duplicating info.

24_clustering.Rmd
```diff
@@ -140,8 +108,7 @@ library(mia)
 library(vegan)
 
 # Load experimental data
-data(peerj13075)
-(tse <- peerj13075)
+tse <- peerj13075
```
antagomir, line 111: I would use the enterotype dataset as an example with all sample clustering examples.

BananaCancer:
This PR needs [miaViz](https://github.com/microbiome/miaViz/pull/80) to be updated in order to work

Issue #288, opened 2023-06-20
MAE examples to improve


antagomir:
The `getWithColData` function extracts MAE components with colData:
`tse <- getWithColData(mae, "microbiota")`

This should be used in MAE examples instead of the two-step process:

```
tse <- mae[["myexp"]]
colData(tse) <- colData(mae)
```

Issue #125, opened 2022-01-08
Example on agglomerateByPrevalence


antagomir:
Add example on `agglomerateByPrevalence`

antagomir:
.. to show how the rare taxa can be aggregated into a single group, instead of filtering them out (which is sometimes done but will then ignore their total combined abundance and bias the relative abundance estimates)

antagomir:
Chapter 6.3 in OMA https://microbiome.github.io/OMA/taxonomic-information.html#data-agglomeration could be a suitable place.

PR #301, opened 2023-07-06
#125, #288

himmil:


11_taxonomic_information.Rmd
```diff
@@ -164,7 +164,16 @@ assay(altExp(tse, "Family"), "counts")[1:5,1:7]
 ```
 
 `altExpNames` now consists of `Family` level data. This can be extended to use 
-any taxonomic level listed in `mia::taxonomyRanks(tse)`.   
+any taxonomic level listed in `mia::taxonomyRanks(tse)`.
+
+Rare taxa can also be aggregated into a single group "Other" instead of filtering them out. A suitable function for this is `agglomerateByPrevance`.
+
+```{r}
+altExp(tse, "Family_prevalence") <- agglomerateByPrevalence(tse, rank = "Family", other_label = "Other")
```
antagomir, line 172: Two points:

1. would be good to explicitly show how and which abundance and prevalence thresholds are used here to define/quantify "prevalence" i.e. just make those arguments explicit in the call 
2. to consider: this example uses altExp, which is a more complex structure within TreeSE; the example could be simplified and made more focused on the core topic of agglomeration by prevalence if we would not use the altExp here. But I have mixed feelings since in a way it is good to make users familiar with all these possibilities.. anyway I would consider removing the altExp part and just operating with the main tse directly(?). Family is also rather high level, usually the number of rare taxa is much higher on the ASV/species level and the need for agglomeration greater.

23_multi-assay_analyses.Rmd
```diff
@@ -71,17 +71,17 @@ mae[[1]] <- mae[[1]][!is.na(rowData(mae[[1]])$Phylum), ]
 rowData(mae[[1]]) <- DataFrame(apply(rowData(mae[[1]]), 2, 
                                      str_remove, pattern = "._[0-9]__"))
 # Microbiome data
-mae[[1]]
+getWithColData(mae, 1)
```
antagomir, line 74: Instead of indices, use the experiment names "microbiota", "metabolites", "biomarkers".. for clarity. 

Before that you can show how to check what experiments there are, with `experiments(mae)`

11_taxonomic_information.Rmd
himmil: 2. Due to the other examples in the same chapter, I'd keep the altExp in this case as well

antagomir:
can we resolve the issues and close this soon?

antagomir:
Hmm is this ready?

antagomir:
Always remember to close the issues after merging (if appropriate).

PR #316, opened 2023-07-29
Streamline biom import section

RiboRings:
Hi! I tried to make the biom import instruction clearer and put example datasets in microbiome/data repository

### DIFF
--- a/04_containers.Rmd
+++ b/04_containers.Rmd
@@ -276,6 +276,12 @@ data("GlobalPatterns", package="mia")
 GlobalPatterns
 ```
 
+#### Tengeler2020 {#tengeler-desc}
+
+Tengeler2020 is derived from a randomised blinded study on the effects of gut
+microbiome on attention-deficit/hyperactivity disorder (ADHD) in humanised mice
+[@tengeler2020gut]. The dataset is briefly presented in
+[these slides](https://microbiome.github.io/outreach/tengeler2020_presentation.html).
 
 #### HintikkaXOData {#hintikka-desc}
 
@@ -616,17 +622,12 @@ Specific import functions are provided for:
 This example shows how [Biom files](https://biom-format.org/) are imported into
 a `TreeSummarizedExperiment` object.
 
-The data is from following publication: 
-Tengeler AC _et al._ (2020) [**Gut microbiota from persons with
-attention-deficit/hyperactivity disorder affects the brain in
-mice**](https://doi.org/10.1186/s40168-020-00816-x). 
-
-The dataset consists of 3 files:
+The dataset consists of 3 files, which can be fetched or downloaded from
+[this repository](https://github.com/microbiome/data/tree/main/Tengeler2020):
 
 -   biom file: abundance table and taxonomy information
 -   csv file: sample metadata
 -   tree file: phylogenetic tree
-
 
 Store the data in your desired local directory (for instance, _data/_ under the
 working directory), and define source file paths
@@ -637,16 +638,21 @@ sample_meta_file_path <- "data/Mapping_file_ADHD_aggregated.csv"
 tree_file_path <- "data/Data_humanization_phylo_aggregation.tre"
 ```  
 
-Now we can load the biom data into a SummarizedExperiment (SE) object.
+Now we can read in the biom file and convert it into a TreeSE object. In addition, we retrieve the rank names from the prefixes of the feature names and then remove them with the `rankFromPrefix` and `removeTaxaPrefixes` optional arguments.
 
 ```{r}
 library(mia)
 
-# Imports the data
-se <- loadFromBiom(biom_file_path)
+# Read in biom file
+biom_obj <- biomformat::read_biom(biom_file_path)
+
+# Convert biom to TreeSE
+tse <- makeTreeSEFromBiom(biom_obj,
+                          rankFromPrefix = TRUE,
+                          removeTaxaPrefixes = TRUE)
 
 # Check
-se
+tse
 ```  
 
 The `assays` slot includes a list of abundance tables. The imported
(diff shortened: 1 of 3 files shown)


04_containers.Rmd
```diff
@@ -637,24 +638,29 @@ sample_meta_file_path <- "data/Mapping_file_ADHD_aggregated.csv"
...
 
-# Imports the data
-se <- loadFromBiom(biom_file_path)
+# Read in biom file
+biom_obj <- biomformat::read_biom(biom_file_path)
+
+# Convert biom to TreeSE
+tse <- makeTreeSEFromBiom(biom_obj,
+                          rankFromPrefix = TRUE,
+                          removeTaxaPrefixes = TRUE)
```
antagomir, line 652: Why don't we use `loadFromBiom` as it is shorter than calling two different functions?

04_containers.Rmd
```diff
@@ -637,24 +638,29 @@ sample_meta_file_path <- "data/Mapping_file_ADHD_aggregated.csv"
...
+tse
 ```  
 
 The `assays` slot includes a list of abundance tables. The imported
 abundance table is named as "counts".  Let us inspect only the first
 cols and rows.
 
 ```{r}
-assays(se)$counts[1:3, 1:3]
+assays(tse)$counts[1:3, 1:3]
```
antagomir, line 663: I think this should be assay(tse) and not assays(tse)?

04_containers.Rmd
```diff
@@ -663,95 +669,56 @@ shows just the beginning of the data table for an overview.
 `knitr::kable()` is for printing the information more nicely.
 
 ```{r}
-head(rowData(se))
+head(rowData(t6se))
```
antagomir, line 672: typo

04_containers.Rmd
```diff
@@ -663,95 +669,56 @@ shows just the beginning of the data table for an overview.
...
-rownames(sample_meta) <- sample_meta[,1]
-
-# Delete column that included sample names
-sample_meta[,1] <- NULL
+# CSV file without headers and rownames in the first column
+sample_meta <- read.table(sample_meta_file_path,
+                          sep = ",", header = FALSE, row.names = 1)
 
 # We can add headers
 colnames(sample_meta) <- c("patient_status", "cohort", "patient_status_vs_cohort", "sample_name")
```
antagomir, line 706: Why not read the metadata file with `DataFrame(read.csv(sample_meta_file_path, sep = ",", header = TRUE))` as this will read the headers automatically (no need to add afterwards; although it must be made sure that examples will work with the original names).

04_containers.Rmd
```diff
@@ -766,10 +733,18 @@ tse
 
 Now `rowTree` includes a phylogenetic tree:
 
-```{r, eval=FALSE}
+```{r, eval = FALSE}
```
antagomir, line 736: in chunks it is recommended style to have no spaces around "="

04_containers.Rmd
```diff
@@ -766,10 +733,18 @@ tse
...
 head(rowTree(tse))
 ```
 
+Finally, we save the complete TreeSE object as an RDS file, so that we can
+readily import and use it in the future.
+
+```{r, eval = FALSE}
+# SAVE as RDS file
+saveRDS(tse, file = "tse.rds")
+```
```
antagomir, line 746: Do we really need to show this, unless we ourselves need it for book examples (which is not the case here with eval=FALSE). -> I suggest to remove lines 740-746

antagomir: This is sort of standard R knowledge, we don't have to teach everything.

book.bib
```diff
@@ -1413,4 +1413,14 @@ @Article{ZIBB2018
   langid = {english}
 }
 
+@article{tengeler2020gut,
```
antagomir, line 1416: I suggest "Tengeler2020" as the bibtex key, to keep our reference style consistent (thus easier to remember)

book.bib
```diff
@@ -1413,4 +1413,14 @@ @Article{ZIBB2018
...
+  title={Gut microbiota from persons with attention-deficit/hyperactivity disorder affects the brain in mice},
+  author={Tengeler, Anouk C and Dam, Sarita A and Wiesmann, Maximilian and Naaijen, Jilly and Van Bodegom, Miranda and Belzer, Clara and Dederen, Pieter J and Verweij, Vivienne and Franke, Barbara and Kozicz, Tamas and others},
```
antagomir, line 1418: can we list all author names here?

book.bib
```diff
@@ -1413,4 +1413,14 @@ @Article{ZIBB2018
...
+  journal={Microbiome},
+  volume={8},
+  pages={1--14},
+  year={2020},
+  publisher={Springer}
```
antagomir, line 1423: would be good to add doi as well. It will create automated link in rendering with certain styles

antagomir (changes requested):

04_containers.Rmd
RiboRings: The colnames specified in line 706 are easier to understand than the original names, so I guess that's why it was done this way.

antagomir: Yes that is the reason. Adding headers is not necessary here, although it might be convenient. 
 

If we keep that, then it could be pedagogically useful to extend the example a bit and first show the "difficult" names, then explain that they are not desirable and show how the user can optionally change those initial names that were already read from the file automatically if they choose so. 

We can do that. It is about finding the optimal balance and focus on teaching the new microbiome methods vs. teaching how to use R for standard data polishing.

antagomir:
up?

antagomir:
just merge when good

PR #320, opened 2023-08-03
Add an `if` condition to check whether we have to remove duplicates in microbiome data

artur-sannikov:
In `23_multi-assay_analyses.R`, on [line 111](https://github.com/microbiome/OMA/blob/master/R/23_multi-assay_analyses.R#L111), duplicates are removed from microbiome data. That is not always necessary. For example, in *this case*, there are no duplicates. I added an `if` condition to check whether we have to remove them. If TRUE, then we remove them.


23_multi-assay_analyses.Rmd
```diff
@@ -179,8 +179,13 @@ colData(mae)$Diet <- ifelse(colData(mae)$Diet == "High-fat" |
                               colData(mae)$Diet == "High-fat + XOS", 
                             "High-fat", "Low-fat")
 
-# Removing duplicates at the microbiome data
-# which are also in form e.g. "Ambiguous" and "uncultured" taxa
+# Remove duplicates from microbiome data if present
+if (any(duplicated(rownames(assay(mae[[1]]))))) {
+  mae[[1]] <-
+    mae[[1]][!duplicated(rownames(assay(mae[[1]]))), ]
+}
```
antagomir, line 186: Would this work instead (and "if" is not necessarily needed since nothing would change if there are no duplicates):

`rownames(mae[[1]]) <- getTaxonomyLabels(mae[[1]], make_unique=TRUE)`


I think `rownames(mae[[1]])` should be identical to the longer `rownames(assay(mae[[1]]))`.

Might less error prone to call mae by name, like `mae[["microbiota"]]`?

-> Update the PR accordingly?

antagomir (changes requested):

23_multi-assay_analyses.Rmd
artur-sannikov: Yes, we can do that. It's then just a shorter version of `mae[[1]] <- mae[[1]][!duplicated(rownames(assay(mae[[1]]))), ]` that was before. I introduced `if` because sometimes we do not have any duplicates.

Now, the purpose changes: we always opt for removing duplicates no matter if they exist or not. What do you think?

antagomir: Yep I agree. No strong opinions here. If there are no duplicates then that line does not really do anything, so it does not matter if it is there or not. 

Doing "if" makes the example more complicated. However, adding the line is also more complicated than necessary in those cases where duplicates do not exist. 

We could do it according to the specific data set. If the present demo data set in the example does not need this then perhaps it can be omitted. In that case it might still be helpful add a brief note in the text instead that if the user encounters duplicates, they can have a look at the `getTaxonomyLabels` function and its `make_unique` argument?

artur-sannikov: Yes, I agree on making a note and removing the code entirely. This will also make the reader reason about their data, rather than blindly copy-pasting the code, so this change can also have some educational value. If you agree, I can make this change and change PR.

antagomir: Yes.

23_multi-assay_analyses.Rmd
```diff
@@ -171,6 +171,9 @@ The `mae` object could be used straight to create the MOFA model. Yet,
 we transform our assays since the model assumes normality per
 default. We can also use Poisson or Bernoulli distributions among others.
 
+Note that duplicates, such as "uncultured", might appear when aggregating the microbiome data by a taxonomic rank. To check for duplicates, run `any(duplicated(rownames(assay(mae[[1]]))))`. If it returns `TRUE`, then the
```
antagomir, line 174: Unneccessarily long, this could be 

`any(duplicated(rownames(mae[[1]])))`

antagomir (approved):
super

Issue #322, opened 2023-08-08
Change black color of significant p-values on heatmap to a more constrasting color


artur-sannikov:
In [Cross-correlation](https://microbiome.github.io/OMA/multi-assay-analyses.html#cross-correlation) black "X" are barely visible on negative blue correlations. I suggest using a green shade which are contrasting with both colors.

For example, [#1dff00](https://www.color-hex.com/color/1dff00) is a good option. See example below.

![fig-hintikka-family-spearman-heatmap-1](https://github.com/microbiome/OMA/assets/40318410/b1c49cb8-7aa9-4c55-9e53-cf70d7084a3a)

antagomir:
Yes, nice! Welcome. Also the cross size could be increased on the same go.

PR #324, opened 2023-08-08
Change color and size of crosses on heatmap

artur-sannikov:
Following #322, I change the color of crosses and increase their size from 8 to 10. Now it looks much better.


antagomir:
Great! Also always close the related issue/s after the merge.

PR #335, opened 2023-08-22
Improve RDA and DAA chapters

RiboRings:
Hi! Here I simplified the RDA section and clarified the DAA chapter. In particular:
- permanova testing is internally executed by runPCA, so no need to do it manually
- clarified complicated output of DAA with confounding variables
- added link to benchmarking repository

### DIFF
--- a/95_resources.Rmd
+++ b/95_resources.Rmd
@@ -28,7 +28,7 @@
 [NGS Analysis Basics](http://girke.bioinformatics.ucr.edu/GEN242/tutorials/rsequences/rsequences/) provides a walk-through of the above-mentioned features with detailed examples.
 
 
-### Alternative containers for microbiome data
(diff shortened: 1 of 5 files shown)


95_resources.Rmd
```diff
@@ -28,7 +28,7 @@
 [NGS Analysis Basics](http://girke.bioinformatics.ucr.edu/GEN242/tutorials/rsequences/rsequences/) provides a walk-through of the above-mentioned features with detailed examples.
 
 
-### Alternative containers for microbiome data
+### Phyloseq: an alternative container for microbiome data
```
antagomir, line 31: It is "phyloseq", not "Phyloseq" (same holds when sentence starts if it is a name that should be written with lowercase)

antagomir (approved):
Good!

PR #343, opened 2023-09-11
Add plotRDA function to beta diversity chapter

RiboRings:
Hi! I implemented the new function [plotRDA](https://github.com/microbiome/miaViz/pull/87) from miaViz in the RDA section, so now it is easier to follow.

This PR also fixes deployment bugs due to missing packages, so when merged check will pass also in other PRs.


20_beta_diversity.Rmd
```diff
@@ -135,31 +120,40 @@ p <- p + labs(x = paste("PCoA 1 (", round(100 * rel_eig[[1]], 1), "%", ")", sep
...
+between them or select the most suitable one to visualize beta diversity in the
+light of the research question.
 
 ```{r plot-mds-nmds-comparison, fig.cap = "Comparison of MDS and NMDS plots based on the Bray-Curtis or euclidean distances on the GlobalPattern dataset."}
+# Run MDS on counts assay with Euclidean distances
 tse <- runMDS(tse,
               FUN = vegan::vegdist,
               name = "MDS_euclidean",
               method = "euclidean",
               assay.type = "counts")
```
TuomasBorman, line 134: This is not standard way to do as far as I know. PCoA with Wuclidean distances is PCA. You could use bray curtis dissimilarities with relative abundances which is a common choice

20_beta_diversity.Rmd
```diff
@@ -135,31 +120,40 @@ p <- p + labs(x = paste("PCoA 1 (", round(100 * rel_eig[[1]], 1), "%", ")", sep
...
 
+# Run NMDS on counts assay with Bray-Curtis distances
 tse <- runNMDS(tse,
                FUN = vegan::vegdist,
                name = "NMDS_BC")
```
TuomasBorman, line 140: Does this really use bray-curtis or other? It would be good to specify with an argument.

20_beta_diversity.Rmd
TuomasBorman: clr + aitchison is also

TuomasBorman: MDS == PCoA

20_beta_diversity.Rmd
```diff
@@ -75,16 +58,22 @@ demonstration we will analyse beta diversity in GlobalPatterns, and observe the
...
+# Beta diversity metrics like Bray-Curtis are often applied to relabundances
 tse <- transformAssay(tse,
+                      assay.type = "counts",
                       method = "relabundance")
 
+# Other metrics like Aitchison to clr-transformed data
+tse <- transformAssay(tse,
+                      assay.type = "relabundance",
+                      method = "clr",
+                      pseudocount = 1)
```
TuomasBorman, line 75: This should not affect (?) since every value get the same pseudocount, but i the documentation we suggest to use smallest positive value --> min(mat[mat > 0])

--> could we add `pseudocount = TRUE` option for this, and give a message on what pseudocount was used? (adding smallest positive value requires user to fetch first the assay which is not optimal...)

20_beta_diversity.Rmd
```diff
@@ -135,32 +127,54 @@ p <- p + labs(x = paste("PCoA 1 (", round(100 * rel_eig[[1]], 1), "%", ")", sep
...
+
+# Run MDS on clr assay with Aitchison distances
 tse <- runMDS(tse,
               FUN = vegan::vegdist,
-              name = "MDS_euclidean",
-              method = "euclidean",
-              assay.type = "counts")
+              method = "aitchison",
+              assay.type = "clr",
+              pseudocount = 1,
```
TuomasBorman, line 148: Why here is pseudocount, probably mistake

20_beta_diversity.Rmd
```diff
@@ -135,32 +127,54 @@ p <- p + labs(x = paste("PCoA 1 (", round(100 * rel_eig[[1]], 1), "%", ")", sep
...
+              name = "MDS_aitchison")
 
+# Run NMDS on clr assay with Aitchison distances
 tse <- runNMDS(tse,
                FUN = vegan::vegdist,
-               name = "NMDS_BC")
+               method = "aitchison",
+               assay.type = "clr",
+               pseudocount = 1,
```
TuomasBorman, line 156: same here

20_beta_diversity.Rmd
```diff
@@ -282,10 +299,10 @@ them. The result shows how much each covariate affects beta diversity. The table
...
 methods.
 
-|                           | supervised ordination  | unsupervised ordination  |
-|:-------------------------:|:----------------------:|:------------------------:|
-| Euclidean distance        | RDA                    | PCA                      |
-| non-Euclidean distance    | dbRDA                  | PCoA                     |
+|                          | supervised ordination  | unsupervised ordination  |
+|:------------------------:|:----------------------:|:------------------------:|
+| Euclidean distance       | RDA                    | PCA                      |
+| non-Euclidean distance   | dbRDA                  | PCoA                     |
```
TuomasBorman, line 306: What is MDS? What is NMDS? What is UMAP? These info could be added

20_beta_diversity.Rmd
RiboRings: Both here and below, an error arises if pseudocount is 0:
```
Error: 'clr' cannot be used with non-positive data: use pseudocount > 5.19440001749332e-05
```

RiboRings: The way I think of it is the pseudocount in `transformAssay(method = "clr")` ensures that the input matrix is positive, but doesn't guarantee that the output matrix is completely positive

TuomasBorman: Ahh ok, aitchison dissimilarity is CLR + Euclidean distance. https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5695134/

I did not know that it is possible to calculate CLR in vegdist. Also I remembered incorrectly Aithchison dissimilarity.

--> so the method should be here Euclidean distance

20_beta_diversity.Rmd
antagomir: Good question. Indeed 1 is problematic for relabundances for instance.  I think the idea has been that the users must explicitly do this. But I am not sure how necessary that is really.

Ok we can go with automation.

antagomir: Often min/2 is used. To distinguish from min. 

However this would work better for relabundances than for integer counts.. so perhaps just min as the default pseudocount..

20_beta_diversity.Rmd
antagomir: pseudocount doesn't make sense in the context of CLR. It is added in order to calculate CLR (before CLR) if the data has zeroes - but never on top of CLR (after CLR).

It is possible to calculate CLR and rCLR transformations with vegan::decostand and their corresponding (r)Aitchison distances in vegan::vegdist, after my last years PR in vegan: 
https://github.com/vegandevs/vegan/pulls?q=is%3Apr+author%3Aantagomir+

Zero imputation procedure for the rCLR / rAitchison is yet to be added.

20_beta_diversity.Rmd
antagomir: Yep. Aitchison applies on counts or relabundances because it does CLR + Euclid automatically. 

Or you can do them separately as well.. but not both at the same time.

Which brings the question back a bit, should we aim to warn users when they do strange combinations? I am not sure, education and good example material could also take us a long way, it is difficult to imagine all unintended use cases.

RiboRings: Conventional combinations could be explained in OMA so warning could be avoided?

20_beta_diversity.Rmd
RiboRings: Aren't they all subtypes of PCoA?

RiboRings:
[mia PR](https://github.com/microbiome/mia/pull/449) to add TRUE/FALSE options for pseudocount

20_beta_diversity.Rmd
antagomir: We could mention some of the most typical combinations, yes. These include at least:

- relative abundance transformation + Bray-Curtis dissimilarity index  (for relative microbiota profiling)  
- original counts + Bray-Curtis dissimilarity index (for quantitative microbiota profiling) 
- CLR-transformation + Euclidean distance (= this combination is called Aitchison distance)
- Robust CLR-transformation + Euclidean distance (= this combination is called robust Aitchison distance)

antagomir: Perhaps also worth mentioning: Jaccard index; you can become more similar to something without becoming less similar to something else..

20_beta_diversity.Rmd
antagomir: PCA is a special case of PCoA (aka MDS) with Euclidean distances.

I am not sure if NMDS is exactly a special case of PCoA/MDS. It is closely related conceptually (focusing on order of distances rather than distances themselves) but saying that it is a special case would require more in-depth check. I don't  think MDS would directly reduce the NMDS but happy to be corrected if mistaken.

UMAP is a different, although related method. It is not a special case of any of the above.

RiboRings:
This PR is ready. If everything looks good I would suggest to merge it to debug the builds and come back to smaller fixes on another PR.

20_beta_diversity.Rmd
```diff
@@ -135,32 +136,54 @@ p <- p + labs(x = paste("PCoA 1 (", round(100 * rel_eig[[1]], 1), "%", ")", sep
 p
 ```
 
-With additional tools from the ggplot2 package, ordination methods can be 
-compared to find similarities between them or select the most suitable one to
-visualize beta diversity in the light of the research question.
+A few combinations of beta diversity metrics and assay types are tipically
```
TuomasBorman, line 139: tipically

TuomasBorman (approved):
Very nice

20_beta_diversity.Rmd
```diff
@@ -16,18 +16,27 @@ knitr::opts_chunk$set(
...
 
+| Method description         | Assay type          | Beta diversity metric |
+|:--------------------------:|:-------------------:|:---------------------:|
+| Quantitative profiling     | Absolute counts     | Bray-Curtis           | 
+| Relative profiling         | Relative abundances | Bray-Curtis           |
+| Aitchison distance         | clr                 | Euclidean             |
+| Robust Aitchison distance  | rclr                | Euclidean             |
+| Present/Absence similarity | Absolute counts     | Jaccard               |
+| Phylogenetic distance      | Absolute counts     | Unifrac               |
+
```
TuomasBorman, line 39: @antagomir Can you quickly check if these are correct

antagomir: Present/Absence -> Presence/Absence
-> and here both absolute or relative counts are ok, they will yield the same result

(r)Aitchison can also be done with "aitchison" and "no transformation" (i.e. the clr is already part of the distance itself). Both are available in mia, and should give identical result. It might be worth mentioning (comment field?) that both options are available.

Phylogenetic distance: hmm at least I would rarify the data; although that might be useful also for Bray-Curtis with relative abundances (albeit controversial..)

PR #346, opened 2023-09-19
update 30_differential_abundance.Rmd

Elina297:
For GUniFrac v1.8 the argument meta.dat has to be removed


RiboRings:
Thank you for the PR! I would ask you to open again this pull request not towards the master branch, but towards the [next_stage branch](https://github.com/microbiome/OMA/tree/next_stage), so that we fix all bugs in that branch and then push to master.

To do this, you need to do:
1. `git fetch; git pull` to update your fork and then
2. `git checkout next_stage` to move to the next_stage branch
3. then you can make the change, push to the next_stage branch on your fork in GitHub (you can just do `git push`, which pushes to the corresponding origin/branch of the current local branch)
4. and open a PR from Elina297:next_stage to OMA:next_stage.

Don't hesitate to ask if you have questions :)

PR #348, opened 2023-09-25
Update 30_differential_abundance

Elina297:
For GUniFrac v1.8 the argument meta.data has to be removed in ZicoSeq.plot


RiboRings:
Nice thanks!

PR #349, opened 2023-09-28
Update the installation of packages

Elina297:
I added in oma_packages.csv the packages that were in DESCRIPTION file.
I added in install_packages.R the MicrobiotaProcess package with "devtools::install_github" because it didn't work when it was in the list of oma_packages.csv.


RiboRings:
Thank you! A few points:
- Did you check if the book can be deployed without MicrobiotaProcess? Because to me it seems that this package is only mentioned but never used in the OMA code.
- Similarly, the ggord package is no longer needed and can be removed wherever it appears
- Check if other packages are not needed for OMA deployment (you can do this by checking if the package under question is used in any of the chapters, and if it doesn't appear uninstall the package from your RStudio and try to deploy OMA)

install_packages.R
```diff
@@ -103,6 +103,7 @@ for(i in seq_along(chapter_pkgs)) {
 # Github packages
 devtools::install_github("microbiome/miaTime")
 devtools::install_github("fawda123/ggord")
```
RiboRings, line 105: no longer needed

install_packages.R
```diff
@@ -103,6 +103,7 @@ for(i in seq_along(chapter_pkgs)) {
...
+devtools::install_github("YuLab-SMU/MicrobiotaProcess")
```
RiboRings, line 106: check if this is really needed

RiboRings (changes requested):
Thank you! A few points:

- Did you check if the book can be deployed without MicrobiotaProcess? Because to me it seems that this package is only mentioned but never used in the OMA code.
- Similarly, the ggord package is no longer needed and can be removed wherever it appears
- Check if other packages are not needed for OMA deployment (you can do this by checking if the package under question is used in any of the chapters, and if it doesn't appear uninstall the package from your RStudio and try to deploy OMA)

antagomir:
@Elina297 can you have a look?

Elina297:
I made a list of packages that where not needed for me to deploy OMA, but I know that in this list there are some important dependencies that should be kept (at least you told me that MultiAssayExperiment is needed).  As I don't know well the packages and dependencies from this book, is there somewhere where I can find the information about what should I let in the installation script even if it worked without that for me ?
Or should I just make a PR removing all those ones and then you check and tell me to add back the ones that are needed ?

Elina297:
I made a list of packages that where not needed for me to deploy OMA, but I know that in this list there are some important dependencies that should be kept (at least you told me that MultiAssayExperiment is needed).  As I don't know well the packages and dependencies from this book, is there somewhere where I can find the information about what should I let in the installation script even if it worked without that for me ?
Or should I just make a PR removing all those ones and then you check and tell me to add back the ones that are needed ?

antagomir:
Ok thanks. 

I can now see. The package listing in "oma_packages.csv" is automatically generated in _build.R_

This includes then all packages that _have been loaded_ when the OMA Rmd source files have been run.

This means that all of them are being loaded when running OMA examples. There is no unnecessary packages since these are automatically loaded when running OMA Rmd files (as one can check from build.R file).

This list is meant to be a comprehensive listing of all packages that are in some way dependencies for the OMA examples. It can be used for comprehensive setup and testing of all necessary dependencies.

My suggestion is to keep this automated list _as is_. This way we can test if there are any packages that cause issues in any given system.

antagomir:
If OMA DESCRIPTION file contains unnecessary packages those can be removed. I am not sure how to check if there are some unnecessary ones and if there is no clear problem associated with this, then perhaps it is not necessary right now to bother about it.

Elina297:
I think I already found some unnecessary packages in DESCRIPTION. If I can find back my data from my laptop I'll make a PR for that.

antagomir:
Ok shall we close this PR then?

Elina297:
I think yes

PR #356, opened 2023-10-16
Add benchdamic to DAA chapter

mcalgaro93:
As stated in the DAA chapter, it is highly recommended to perform DAA with multiple methods to determine whether the findings can be reproduced by different approaches. 

I've added a couple of sentences about `benchdamic` which is a Bioconductor package enabling users to assess the performance of DAA methods in their individual datasets. 

Thank you for this amazing book,
Matteo Calgaro


antagomir (approved):

antagomir:
Great, thanks a lot! I also specifically like the approach of not replicating code that is already well done elsewhere but instead linking to such external resources.

More PRs welcome! We will close this after the checks will pass.

TuomasBorman (approved):

Issue #232, opened 2023-03-12
Subsection on "How to contribute"


antagomir:
The current [OMA README](https://github.com/microbiome/OMA/blob/master/README.md) has brief instructions on how to contribute to OMA. 

It has been suggested to add/move instructions to a dedicated subchapter in OMA itself, for instance under the "Acknowledgments" chapter.

antagomir:
Hmm perhaps still good to have this in README. 

Let us add a small subchapter "How to contribute" in the "Developers" section. That subchapter would simply mention that the instructions can be found from [README](https://github.com/microbiome/OMA/blob/master/README.md) and link to that README.

antagomir:
will assign this to EC - HL can support.

Issue #124, opened 2022-01-08
perSampleDominantTaxa


antagomir:
Add dominant taxa PCoA visualization with `perSampleDominantTaxa` to beta diversity chapter.

Also show example on `countDominantTaxa`

Include citation to [Salosensaari et al. 2021](https://doi.org/10.1038/s41467-021-22962-y)

antagomir:
We may have this already. Check, and close the issue if it has been done.

himmil:
`countDominantTaxa` and `addPerSampleDominantTaxa` -> `getTopTaxa` don't find identical results for dominant taxa for each sample. What is the functionality of `countDominantTaxa` and why the results differ?

antagomir:
Ok - here:

The `countDominantTaxa` generates a table (tibble) that lists 
1. the dominant taxa
2.  the count of samples where that table is dominant (ie. it is the most abundant one; note that that multiple taxa could have the same maximum abundance level at the same time in principle); for instance: `sum(assay(tse, "counts")["331820",] == apply(assay(tse, "counts"), 2, max))`
3. the frequency of the samples that are dominated by that taxon: e.g. `sum(assay(tse, "counts")["331820",] == apply(assay(tse, "counts"), 2, max)) / ncol(tse)` and somewhat confusingly this is provided as percentage [0,100]
4. the same as (3) but rounded to closest integer and added with "%" so it becomes a character..

I am not sure how much all of that is needed but it seems like a useful information to have the number of samples that are dominated by a particular taxonomic feature. Whether we really need both fields 3 and 4 I am not sure.

I would suggest the following:

1. at least explain this function more clearly in function documentation (PR to mia!); explaining this in OMA is a bit optional, we don't have to explain every single function in OMA
2. change field (3) into scale [0,1] (again, PR would go to mia!)
3. remove field (4), users can easily add this if it is necessary? (again, PR would go to mia!)
4. in the example, show how to complement the tibble with rowData (i.e. instead of the crypctic ASV codes, one can include also for instance Genus information from rowData  in the tibble)?

If the changes are agreed here I think PR could be directly opened to mia.

antagomir:
Would be curious to see if @TuomasBorman has further comments.

TuomasBorman:
Sounds good

1. OK
2. OK
3. Removing or keeping are both ok (I would probably just remove it). Removing --> would be more clear; the value is already there and can be easily modified. Keeping --> could give additional value when printed in report.
4. OK

[...]

PR #357, opened 2023-10-18
#124 #232

himmil:
124: Example of countDominantFeatures
232: How to contribute


20_beta_diversity.Rmd
```diff
@@ -402,6 +402,8 @@ tse_genus <- transformAssay(tse,
 tse_genus <- addPerSampleDominantFeatures(tse_genus,
                                           assay.type = "relabundance",
                                           name = "dominant_taxa")
+# Overview
+countDominantFeatures(tse_genus, rank = "Genus", digits = 3, name = "dominant taxa")
```
antagomir, line 406: we should generally avoid spaces in names -> use e.g. dominant_taxa instead

PR #358, opened 2023-10-18
Added phyloseq vs mia cheatsheet and examples

KatariinaParnanen:


97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+We have counts!
+
+Since TreeSEs can hold many different versions of the OTU table, most commonly either relative abundances or counts, we will need to give our assay a name and list the different types of assays we have. In this example we only have one item 'counts' in the list.
+
+Let's convert the data frame to a matrix and make the list of assays.
+
+```{r}
+# Create TreeSE
+counts <- as.matrix(PHY_otutable) # Convert to a matrix
+assays <- SimpleList(counts = counts)
```
antagomir, line 392: If you would create the abundance table taxtable and sample phenodata in one chunk _before_ phyloseq and TreeSE construction, you could show how to construct either of these objects from those same raw materials.

It might be helpful to add some notes about best practices, if for instance rownames/colnames between elements are expected to match etc.

97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+
+TSE <- TreeSummarizedExperiment(assays = assays, 
+                                colData = PHY_sampledata,
+                                rowData = PHY_taxtable
```
antagomir, line 396: slightly confusing that PHY is somehow needed here. See suggestion above to separate data elements before PHY or TSE construction.

97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+)
+
+```
+Let's check the different assay names we have.
```
antagomir, line 400: Perhaps explain that "assay" corresponds to a statistical transformation of hte data?

And that TreeSE can carry multiple ones, vs. phyloseq only one?

97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+```{r}
+assayNames(TSE)
+```
+
+
+### Handling different OTU table normalizations in Phyloseq vs TreeSE
+
+Adding the assays as a list might seem inconvenient if you only have one type of OTU table (counts in our example), but let's see why it is actually very convenient to be able to hold multiple assays in one data object. 
+
+Here we'll show an exampe of how to add relative abundances and CLR normalized OTU tables to your TSE assays.
```
antagomir, line 411: exampe -> example

97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+# Extract
+df <- colData(TSE) %>% data.frame %>% dplyr::select(matches("shannon"))
+
+```
+
+
+### Calculating beta diversity: ordinate = runMDS
+
+We can calculate PcoA with Bray-Curtis distances in Phyloseq using the ordiate command. This 
+The beta diversity calculation in mia  outputs a TreeSE with a new type of data, reduced dimensions or reducedDim. Here we will use the scater package that runs the PcoA with runMDS. (PcoA and MDS mean the same thing)
```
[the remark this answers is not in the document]
antagomir: (two times)

97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+
+In Phyloseq you would again need to add the dimensions to the sample data if you want to keep them safe with other metadata.
+
+```{r}
+# Run PCoA on the relative abundance data and store in PHY_ord list
+PHY_ord <- ordinate(PHY_relab, method = "PCoA", distance = "bray")
+
+library(scater)
+
+# For the ordination using mia we use runMDDS from the scater package and implement the vegan bray distance calculation
```
antagomir, line 539: MDDS -> MDS

97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+```
+
+Tax glom removes the taxa which have not been assigned to the level given in taxrank by default. NArm = TRUE so we will add the onRankOnly = TRUE which is equivalent to the mergeFeaturesByRank.
+
+```{r}
+altExp(TSE, "Family") <-
+  mergeFeaturesByRank(TSE,
+                      rank = "Family",
+                      onRankOnly = TRUE, # Let's add the equivalent option to NArm =TRUE
+                      agglomerateTree = TRUE)
```
antagomir, line 598: This seems to be the default, so should we leave it out for simplicity?

97_extra_materials.Rmd
```diff
@@ -290,3 +290,324 @@ e <- attr(reducedDim(tse_Genus, "PCoA_BC"), "eig");
...
+
+
+### Cheatsheet
+
+```{r}
+library(knitr)
+df <- data.frame(Functionality=c("Access sample data", "Access tax table", "Access OTU table", "Build data object", "Calculate alpha diversity", "Calculate beta diversity", "Plot ordination", "Subset taxa", "Subset samples", "Aggromerate taxa"), phyloseq=c("sample_data()", "tax_table()", "otu_table()", "phyloseq()", "estimate_richness()", "ordinate()", "plot_ordination()", "subset_taxa()", "subset_samples()", "tax_glom()"), miaTreeSE = c("Index columns", "Index rows", "assays()", "TreeSummarizedExperiment()", "estimateDiversity()", "runMDS()", "plotReducedDim()", "Index rows", "Index columns", "mergeFeaturesByRank()"))
+
+df2 <- data.frame(Data_type=c("OTU table", "Taxonomy table", "Sample data table"),phyloseq=c("otu_table", "tax_table", "sample_data"), TreeSE=c("assay","rowData","colData"))
+kable(df)
```
antagomir, line 610: Would be helpful for later edits if this can written in a way that the methods form pairs, for instance one pair per row. So one can directly see from this code easily what is matching what?

antagomir (commented):
Very nice, thank you!

Some suggestions added.

Feel free to add yourself in "90_acknowledgments.Rmd". Otherwise, the contributor list is also visible at the OMA github page and the book links there too.

antagomir:
Quitting from lines 416-434 [unnamed-chunk-26] (97_extra_materials.Rmd)
Error in `loadNamespace()`:
! there is no package called 'microbiome'

-> Should be added to DESCRIPTION file I guess. Kindly resolve.

TuomasBorman:
Noticed that we have very brief explanation in the book already https://microbiome.github.io/OMA/resources.html#phyloseq-an-alternative-container-for-microbiome-data

The table could be added there, but then you could add also link to this extra material

antagomir:
I am ok with either solution but somehow feel that ch 18 "Extra material" could be more appropriate. To make clear separation between these two systems. Hence my suggestion would be to merge the material from ch 16 to ch 18?

antagomir:
@KatariinaParnanen I just noticed that this was never finalized. GIven that you initiated the PR and we have provided feedback, would you be able to finalize this one?

97_extra_materials.Rmd
KatariinaParnanen: onRankOnly is FALSE by default?

KatariinaParnanen:
I made a new commit with the edits to the Rmd per suggestions. Do I need to do something else also?

antagomir:
Ah.. we have switched to Quarto after this PR was opened.. 

@ake123 could you migrate the material from here into the new files, I think these ones and make a PR?
https://github.com/microbiome/OMA/tree/master/inst/pages

Kindly follow-up on the PR until it is closed.

ake123:
Done!

Issue #360, opened 2023-10-18
DMM examples in OMA


antagomir:
DMM examples in OMA seem unnecessarily complex. See if these can be simplified with the latest updates to DMM methods in mia.

PR #373, opened 2023-10-24
#360

himmil:
DMM example simplifications

Edited also some other written parts in the chapter to make it more clear


24_clustering.Rmd
```diff
@@ -221,97 +196,122 @@ tse <- GlobalPatterns
...
+
+drivers$phyla <- gsub("Phylum:", "", rownames(drivers)) # Clean phylum names
+
+plots <- c()
+for (i in 1:2) {
+  drivers <- drivers[order(drivers[[i]], decreasing = TRUE),]
+  p <- ggplot(head(drivers, 10), aes(x = reorder(head(phyla, 10), + head(drivers[[i]], 10)), y = head(drivers[[i]], 10))) +
+          geom_bar(stat = "identity", fill = "deeppink4", alpha = 0.5) +
+          coord_flip() + labs(title = paste("Top phyla in a group", i)) +
+          theme_light(base_size = 15) + xlab("") + ylab("") + ylim(0, 7)
```
antagomir, line 270: I think xlab & ylab and ylim are not any more recommended; use instead:

labs(x="", y="") and scale_y_continuous(limits=c(0,7))

24_clustering.Rmd
```diff
@@ -221,97 +196,122 @@ tse <- GlobalPatterns
...
+  print(p)
+}
 ```
 
-We can also know the contribution of each taxa to each component
 
-```{r dmm9}
-head(DirichletMultinomial::fitted(bestFit))
+We use `calculateDMNgroup` function to have an overview of the best model. The function groups samples by sample type and returns DMNGroup object that contains a summary.
```
antagomir, line 276: should we mention that "SampleType" is a field in colData?

24_clustering.Rmd
```diff
@@ -221,97 +196,122 @@ tse <- GlobalPatterns
...
+
+```{r dmm_group}
+dmm_group <- calculateDMNgroup(tse_dmm, variable = "SampleType", 
+                               assay.type = "counts", k = 2, 
+                               seed = .Machine$integer.max)
```
antagomir, line 281: why this kind of seed rather than e.g. "123" - the current might be somehow confusing?

24_clustering.Rmd
```diff
@@ -221,97 +196,122 @@ tse <- GlobalPatterns
...
+Mixture weights can be used for having a rough approximation of the cluster size.
 
 ```{r dmm10}
-# Do clr transformation. Pseudocount is added, because data contains zeros.
+DirichletMultinomial::mixturewt(bestFit)
+```
+
+### PCoA with DMM clusters
+
+In this section we show how to calculate principal coordinates for clr transformed abundance data. We use Aitchison distance as a distance metrics that calculates Euclidean distances for clr transformed compositions. In the visualization section, the samples are coloured based on their DMM clusters to demonstrate the distances between samples included in each DMM cluster.
```
antagomir, line 294: I may misunderstand something but DMM assumes count data and it has its own model for clustering. 

It does not use CLR data or Euclidean / Aitchison distances. It should be made clear what transformations & distances refer to PCoA, and which to DMM (and that these are distinct)

24_clustering.Rmd
```diff
@@ -221,97 +196,122 @@ tse <- GlobalPatterns
...
+
+```{r dmm11}
+# pseudocount is added, because data contains zeros
```
antagomir, line 297: "pseudocount is added" -> "add pseudocount" (let us aim to use active tense in these)

24_clustering.Rmd
```diff
@@ -221,97 +196,122 @@ tse <- GlobalPatterns
...
 assay(tse, "pseudo") <- assay(tse, "counts") + 1
 tse <- transformAssay(tse, assay.type = "pseudo", method = "relabundance")
+
+# clr transformation
```
antagomir, line 301: It seems that data transformations are done many times in this chapter. Could we just add the transformed assays once in the beginning of the chapter, then call these by assay name in the later chunks?

24_clustering.Rmd
```diff
@@ -221,97 +196,122 @@ tse <- GlobalPatterns
...
 # Create a data frame from principal coordinates
 euclidean_pcoa_df <- data.frame(pcoa1 = df[, 1], pcoa2 = df[, 2])
 ```
 
-```{r dmm11}
+```{r dmm12}
 # Create a data frame that contains principal coordinates and DMM information
 euclidean_dmm_pcoa_df <- cbind(euclidean_pcoa_df,
-                               dmm_component = colData(tse)$clusters)
+                               dmm_component = colData(tse_dmm)$clusters)
```
antagomir, line 314: Hmm how about using altExp for tse_dmm (instead of creating this new TreeSE object tse_dmm)?

24_clustering.Rmd
himmil: Thanks for the comment, I fixed the text to be more clear. I think also using alternative experiments clarify the distinction in the code.

24_clustering.Rmd
himmil: Hclust and DMM examples all use different data all saved with the same name "tse". Should I change it (either names to be different or datas to be the same)? I also made some editions to transformations

antagomir: Let us use just one name "tse" unless there is specific need to use many at the same time. 

Using just "tse" makes it way easier to copypaste and reuse the code for  new purposes.

antagomir:
Is this now complete or still something coming?

himmil:
It's complete. I'll merge it if you don't have any more comments

antagomir:
yess

Issue #188, opened 2022-08-05
Consider Quarto as the basis for the final book version


antagomir:
Quarto appears promising, and especially good feedback on books:
https://quarto.org/docs/gallery/

antagomir:
This has been done with Quarto, for instance:
https://jjallaire.github.io/hopr/

TuomasBorman:
I agree, I have been planning to start using Quarto instead of Rmarkdown in my projects. --> we could convert these Rmd files into Quarto (and start recommending Quarto (for example, one exercise was about creating Rmd file...)

There should be a command that converts automatically Rmd to Quarto if I remember correctly

antagomir:
Yes.

And in many cases Rmd files are readily usable without changes; although it might then be informative to change the file name to `.qmd`.

There may be some issues in converting the book. For instance, the packages seem to be loaded per chapter, one has to check that the necessary packages are load per each chapter. This will probably just improve the usability of the material.

TuomasBorman:
Also, consider using [BiocBook](https://github.com/js2264/BiocBook)

ake123:
Here is an error 
```r
Quitting from lines 217-246 [taxa_clustering] (11_taxonomic_information.qmd)
Error:
! unused argument (dist.fun = function (x, method = "euclidean", diag = FALSE, upper = FALSE, p = 2) 
{
    if (!is.na(pmatch(method, "euclidian"))) method <- "euclidean"
    METHODS <- c("euclidean", "maximum", "manhattan", "canberra", "binary", "minkowski")
    method <- pmatch(method, METHODS)
    if (is.na(method)) stop("invalid distance method")
    if (method == -1) stop("ambiguous distance method")
    x <- as.matrix(x)
    N <- nrow(x)
    attrs <- if (method == 6) list(Size = N, Labels = dimnames(x)[[1]], Diag = diag, Upper = upper, method = METHODS[method], p = p, call = match.call(), class = "dist") else list(Size = N, Labels = dimnames(x)[[1]], Diag = diag, Upper = upper, method = METHODS[method], call = match.call(), class = "dist")
    .Call(C_Cdist, x, method, attrs, p)
})
Backtrace:
 1. mia::cluster(...)
 2. mia::cluster(...)
 4. bluster::clusterRows(mat, BLUSPARAM, full)
 5. bluster:::.cut_hierarchical(hcl, dst, BLUSPARAM)
 6. base::do.call(fun, c(list(hcl), args))

```
[...]

PR #374, opened 2023-10-25
Quarto conversion 

ake123:
Consider Quarto as the basis for the final book version #188


antagomir (approved):
OK to me if the book renders and publishes correctly, and README has up-to-date instructions on how to render & view locally.

ake123:
README file is created  specifically for Quarto conversion and can be found in  `README_quarto_conversion.md` file.

antagomir:
It seems that we will unfortunately have to cancel this PR in the end, because the BiocBook conversion appeared in parallel in #376 .. but let us wait until that is merged, in case there are any changes.

antagomir:
Replaced by #379

PR #377, opened 2023-12-01
Update network chapter and contributors

stefpeschel:
- Reorder network learning chapter to be less nested
- Move network plots to analysis section
- Add comment on closeness centrality
- Add contributors


antagomir (approved):

antagomir:
All other PRs are still against master, so I approve and merge.

TuomasBorman:
> @TuomasBorman do we take the PRs still against the master branch directly? just to check before I merge.

Sorry for delay. --> Yes, I think that we can still use master since we have used it already previously. But after the next Bioc release, let's use devel branch (also in mia packages)

TuomasBorman:
In general, we should follow Bioc suggestions/rules as closely as possible (separate devel branch is one of those)

antagomir:
Yes!

PR #376, opened 2023-11-17
Conversion to BiocBook

js2264:
Hi OMA authors,

Following my issue https://github.com/microbiome/OMA/issues/375, I have initiated some work to convert the OMA book to a `BiocBook` infrastructure [see [here](https://github.com/js2264/BiocBook) for some details].

Few points worth bringing up:

- The pages are now in `.qmd` (quarto) format, and located in `inst/index.qmd` and `inst/pages/*qmd`. You can add any page to the book by creating a new `.qmd` file in `inst/pages/` and reference this file in the `inst/assets/_book.yml` listing.
I have copied and pasted most of the pages content, and made very minor changes (such as fixing the reference calls).
- This PR should actually be merged to a new`devel` branch on the `microbiome/OMA` repo, not the `master` I currently selected (but I cannot directly create this branch in your repo). This would allow automatic detection of the Bioc version to use: a branch named `devel` (or `RELEASE_3_18`) will use bioc `devel` (or bioc `3.18`) to build the book (and deploy it to `gh-pages:docs/3.18`). The `master` branch is not going to trigger automated book building at all.
- If you want to serve the book from the github repository, the Pages settings should be set to serve Pages from the `gh-pages` branch (root folder). An automated redirection sends to `gh-pages:docs/devel`, but can be set up to other versions (if any `RELEASE_*` branch exists). The URL that points to an actual book (once generated) would be https://microbiome.github.io/OMA/docs/devel/ (or https://microbiome.github.io/OMA/docs/RELEASE_*/).
- The repo can be installed with `remotes::install_github("js2264/OMA")`, which sets up all the dependencies from CRAN/Bioc (and `miaTime` from github). Alternatively, a docker image with pre-configured RStudio server is available (`docker pull ghcr.io/js2264/oma:devel`). Once the PR pulled, this would become `microbiome/OMA` of course.

Let me know if there is anything unclear. The automated website (only on Bioc devel for now) is available here: [https://js2264.github.io/OMA/docs/devel/](https://js2264.github.io/OMA/docs/devel/). I have opened this PR to check how changed files would show up. Feel free to postpone the review until ready, I mostly wanted to have it open so my fork (`devel` branch) can be directly compared to your `master`.

PR #378, opened 2023-12-11
Revert "Conversion to BiocBook"

antagomir:
Reverts microbiome/OMA#376


antagomir:
Woops I merged it into devel in the end anyway.. reverting now.

PR #382, opened 2023-12-14
Add redirection to new book

ake123:
Adding redirection to the new book


TuomasBorman (approved):
I believe now we can remove all the other files from gh-pages:root?

ake123:
Yes we can do that

Issue #386, opened 2024-01-17
Remove additional files from gh-pages branch


TuomasBorman:
gh-pages includes additional files from old versions of books. Remove them and clean up the branch so that it includes only necessary files.

PR #388, opened 2024-01-22
Cleaning up the gh-pages

ake123:
Cleaned up the gh-pages branch except the redirection html (index.html)
Remove additional files from gh-pages branch #386


TuomasBorman:
Thanks!

PR #390, opened 2024-02-09
Example on how to add foldable solutions to exercises

RiboRings:
Hi! Before switching to BiocBook we were considering to add foldable solutions to the exercises. Here I added solution chunks to a couple of exercises so if anyone is willing to expand they can use the blueprint used in this PR. For example:

```
#| label: exercise-name-chunk1
#| eval: FALSE
#| code-fold: true
#| code-summary: "Show solution"

# Write solution here
```

Ideally, break down solution chunks into steps same way as the exercises are fragmented.


antagomir (approved):

antagomir:
Nice!

antagomir:
Just merge when happy.

Issue #391, opened 2024-02-13
pseudocount = 1 in clr assay transformation in chapter 5.4


artur-sannikov:
[Chapter 5.4](https://microbiome.github.io/OMA/docs/devel/pages/11_taxonomic_information.html#sec-assay-transform) adds 1 as pseudocount to `relabundance` assay before performing clr transformation.

```R
tse <- transformAssay(x = tse, assay.type = "relabundance", method = "clr", 
                      pseudocount = 1, name = "clr")
```

Does it make sense though? @antagomir in our discussion correctly said that adding 1 to relative abundance does not make much sense because 1 is the largest number a `relabundance` assay can have. So, the correct code should be just `pseudocount=TRUE`.

antagomir:
It does not make sense, the correct code should be `pseudocount=TRUE`. 

Could you fix by PR?

PR #393, opened 2024-02-15
Fix pseudocount for clr transformation from relabundance

artur-sannikov:
Similar to #391, found the same issue in another [chapter](https://microbiome.github.io/OMA/docs/devel/pages/21_microbiome_community.html#sec-composition-heatmap).


antagomir (approved):
Great!

Issue #354, opened 2023-10-05
Add phyloseq vs TreeSE cheatsheet


TuomasBorman:
To help users to convert phyloseq workflows to TreeSE, add info on equivalences (tax_table vs rowData, otu_table vs assays...)

Maybe in picture format...

antagomir:
Under construction in #358 #397

antagomir:
Can be closed? @ake123

ake123:
Yes it can closed!

antagomir:
close if ok

PR #397, opened 2024-04-01
migrate material from PR #358

ake123:
migrate material from PR #358


antagomir:
Check and merge if OK

antagomir:
Also check suggestions from #354

Issue #347, opened 2023-09-20
ggplot2 improvement


antagomir:
Earlier we have had to do `ggplot(as.data.frame(DF))` for `DataFrame` objects.

Now this issue [has been solved](https://github.com/tidyverse/ggplot2/issues/5390).

-> Change OMA examples from `ggplot(as.data.frame(DF))` to just `ggplot(DF)` for `DataFrame` objects.

This requires for the time being that the github version of tidyverse/ggplot2 is being used. Either change that, or wait for the next ggplot2 update in CRAN.

antagomir:
This now seems to work with the latest R release. Simplify all OMA occurrences of ggplot as suggested and check manually that this seems to render correctly. 

To NdG?

PR #403, opened 2024-04-11
Issue 347 fix

Insaynoah:
I changed and simplified the ggplot usages as indicated #347 , everything seems to render correctly.

### DIFF
--- a/inst/pages/04_containers.qmd
+++ b/inst/pages/04_containers.qmd
@@ -374,7 +374,6 @@ Load required packages.
 
 ```{r dada2_2}
 library(mia)
-library(ggplot2)
 library(BiocManager)
 library(Biostrings)
 ```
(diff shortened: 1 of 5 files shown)


antagomir:
squash/merge & close issues if ok.

Insaynoah:
I don't have the rights to do so. Could you grant me them ?

antagomir:
@ake123 could add? 

In general, wait for our explicit approvals before merging.

TuomasBorman:
I think it might be good idea reserve these higher level rights for maintainers. The default branch (devel) should be kept as clean as possible; it should not contain any changes that has not gone through maintainers' review. The risk is that accidents happen, and PR is merged before reviews. Also, for example, just before Bioc release cycle, it is better to not do major changes since if they break something, we are in rush fixing them. --> So it might be better to control devel branch even though it might sometimes mean longer response times.

antagomir:
ok

TuomasBorman:
I noticed that you were missing all the rights --> I added you to miaverse-developers team.

The rights should be high enough. To make review easier, you can create branch to microbiome/OMA directly instead of to your own fork

Issue #361, opened 2023-10-18
Installation scripts & instructions


antagomir:
The current installation script (ch 2.1) installs Bioc versions of mia, miaViz, bluster.

However, OMA uses some stuff from the latest Github devel versions.

Update the current installation instructions to use those Github versions.

During 2024, stabilize OMA so that it is in sync with the latest Bioconductor devel version (and does not need some experimental and unstable Github development versions of the key packages).

antagomir:
To NdG

RiboRings:
Hi! I hope this is the right place to bring up installation issues, but I can also open a separate issue if needed.

When rendering the book locally, I get this warning. Consider adding these dependencies under `Suggests:` in DESCRIPTION or check if some of them can be replaced with base R functionality.

```
> BiocBook::preview(BiocBook::BiocBook('.'))
✖ Some dependencies found in book pages are not listed in DESCRIPTION: 
  • BiocManager
  • stats
  • reshape2
  • devtools
  • NetCoMi
  • SpiecEasi
  • SPRING
  • RColorBrewer
  • factoextra
  • grid
  • cluster
  • dada2
ℹ Consider adding these dependencies to DESCRIPTION
```

RiboRings:
Also, the `kableExtra` package is used in some chapters but is not installed by the `install_packages.R` script:

```
Quitting from lines  at lines 86-90 [unnamed-chunk-6] (10_manipulation.qmd)
Error in `loadNamespace()`:
! there is no package called 'kableExtra'
```

Consider replacing `kableExtra` with a base R solution, or adding it to `install_packages.R`.

antagomir:
I would add it to installation script unless there are particular problems with that. It has useful functionality.

antagomir:
You can modify the oma_packages.csv file at the root folder:
https://github.com/microbiome/OMA

The URL also refers to the Github branch where this is released after your PR

PR #408, opened 2024-04-12
Update install_packages.R

Insaynoah:
I updated the for loop to install the devel version of specified packages as suggested in #361, It should be easy to update, if we ever need to add other devel versions of packages.

There's one minor possible issue which is that if the person already has these packages installed, it doesn't update to the devel version. Should I implement this or does this seem fine?

### DIFF
--- a/install_packages.R
+++ b/install_packages.R
@@ -97,7 +97,17 @@ for(i in seq_along(chapter_pkgs)) {
     message("### CHAPTER: ", i, " ###")
     pkgsAvailable <- installed.packages()[, "Package"]
     pkgsToInstall <- setdiff(chapter_pkgs[[i]], c(pkgsAvailable, pkgs_github))
-    BiocManager::install(pkgsToInstall, update = FALSE, upgrade = FALSE, ask = FALSE, type = pkg_type)
+    #installs the devel versions of mia, miaViz and bluster
+    devel_pkgs <- c("mia", "miaViz", "bluster")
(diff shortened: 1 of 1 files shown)


install_packages.R
```diff
@@ -97,7 +97,17 @@ for(i in seq_along(chapter_pkgs)) {
     message("### CHAPTER: ", i, " ###")
     pkgsAvailable <- installed.packages()[, "Package"]
     pkgsToInstall <- setdiff(chapter_pkgs[[i]], c(pkgsAvailable, pkgs_github))
-    BiocManager::install(pkgsToInstall, update = FALSE, upgrade = FALSE, ask = FALSE, type = pkg_type)
+    #installs the devel versions of mia, miaViz and bluster
+    devel_pkgs <- c("mia", "miaViz", "bluster")
+    if (any(devel_pkgs %in% pkgsToInstall)) {
+        BiocManager::install(devel_pkgs[devel_pkgs %in% pkgsToInstall], update = FALSE, upgrade = FALSE, ask = FALSE, type = pkg_type, version = "devel")
```
TuomasBorman, line 103: As suggested in the issue, the packages should be bleeding-edge versions from GH.

- Github contains the most recent changes.

- Bioconductor devel contains versions that have been gone through testing in GH (there is some days delay in updating).

- Bioconductor release contains versions that are udpating 2 times per year based on Bioconductor devel version.

So instead of installing these from Bioconductor devel, install these packages from GH with remotes. Also add comment that the installation step must be removed before Bioc submission

PR #415, opened 2024-04-16
Issue 361 devel packages

### LINKED ISSUE #361: Installation scripts & instructions

ndegunst:
This should fix #361 where the code installs github versions of the specified packages (mia and miaviz) and the devel version of bluster. It should be relatively easy to add or remove packages as well.


antagomir (approved):

antagomir:
Looks good to me!

PR #423, opened 2024-04-19
Delete double ggsignif library import

artur-sannikov:
In `14_alpha_diversity.qmd`  `library(ggsignif)` was duplicated.


antagomir (approved):

inst/pages/11_taxonomic_information.qmd
```diff
@@ -277,133 +277,40 @@ tse_merged
...
-   this us used to bring features (or samples) to more comparable
-   levels in terms of mean and scale of the values. This can enhance
-   visualization and interpretation of the data
-
-* 'log', 'log2', 'log10' Logarithmic transformations; used e.g. to
-   reduce data skewness; with compositional data the `clr` (or `rclr`)
-   transformation is often preferred.
-   
- * 'hellinger' Hellinger transformation equals to the square root of
-   relative abundances. This ecological transformation can be useful
```
TuomasBorman, line 337: Why there is so many lines deleted?

.github/workflows/rworkflows.yml
```diff
@@ -48,6 +48,6 @@ jobs:
         has_runit: ${{ false }}
         has_latex: ${{ false }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run_docker: ${{ false }}
```
TuomasBorman, line 51: Also docker is enabled

inst/pages/14_alpha_diversity.qmd
```diff
@@ -119,7 +119,6 @@ It adds p-values to plot.
 library(ggsignif)
 library(ggplot2)
 library(patchwork)
-library(ggsignif)
```
TuomasBorman, line 122: This is the contribution explained in this PR

TuomasBorman (changes requested):
Are those other modifications intentional? Many removed lines

inst/pages/11_taxonomic_information.qmd
artur-sannikov: No, they are not. I only focused on the `ggsignif` library.

antagomir: These seem to be deleted in the PR as they are red?

These lines I just added a few days ago, they should remain in the book. If they were also duplicated then the duplicates can be removed of course.

antagomir:
@artur-sannikov - I checked the source branch from which you have opened this PR. 

It indeed has deleted the major chunks from bibliography and some other files, as seen in the PR code change view.

I think this is because some updates had happened for those after you had synced with master branch.

It is essential to not delete all those other things as a result of merging this PR. 

Could you make a fresh clone or sync with the master branch here, and then open a new PR that has only the changes you suggest in this PR, and not the other changes?

artur-sannikov:
> @artur-sannikov - I checked the source branch from which you have opened this PR.
> 
> It indeed has deleted the major chunks from bibliography and some other files, as seen in the PR code change view. You can see this by comparing the master branch and your branch. For instance, Keshavan reference cannot be found from your source branch for this PR but it is here in the original master branch.
> 
> I think this is because some updates had happened for those after you had synced with master branch.
> 
> It is essential to not delete all those other things as a result of merging this PR.
> 
> Could you make a fresh clone or sync with the master branch here, and then open a new PR that has only the changes you suggest in this PR, and not the other changes?

Hi @antagomir, I have synced my master and devel branches. Both my [master ](https://github.com/artur-sannikov/OMA/blob/master/inst/pages/11_taxonomic_information.qmd#L304)and [devel](https://github.com/artur-sannikov/OMA/blob/master/inst/pages/11_taxonomic_information.qmd) branches have deleted lines now.

Should the merging work now?

artur-sannikov:
Yes, you are correct. But they are in my [master branch](https://github.com/artur-sannikov/OMA/blob/master/inst/pages/11_taxonomic_information.qmd#L320). Should a merge of master to devel solve this issue?

antagomir:
Those issues should be solved in the branch that is the source for this PR. It seems to be artur-sannikov:devel

artur-sannikov:
These are not actual issues. My branches are syncronized. I think my previous pull requests were not merged into devel branch, or modified by someone after the merge, so now I have dozens of conflicts. I'm okay with accepting the changes done so far to my local repo.

antagomir:
Was the only purpose of this PR to remove the duplicated occurrence of `library(ggsignif)` in  14_alpha_diversity.qmd ?

Yesterday I rewrote the alpha diversity section and I forgot that there is this PR and solved that duplicate on the same go in #439 

However this PR contains a large number of other (and at least partially unintended) changes, some of which would remove new work. If the intention was to add other changes in addition to duplicate removal, I suggest to take a fresh clone of OMA:devel branch, and make a new PR against it that has only the intended changes.

artur-sannikov:
I copied the current files from the upstream branch to my branch, commited and Git still shows that I have discrepancies between upstream:devel and my fork:devel branches.

antagomir:
In this kind of situations it is often easiest to take a fresh fork of OMA and then add changes in that, rather than get into adding potentially very complicated changes to fix all discrepancies that have occurred along the way.

Another option is to use a syncing mode that replaces your changes with the latest upstream version, and then add new changes on top of that. I do not remember which commands were needed for that, though, but Git tutorials explain the procedures.

Issue #425, opened 2024-04-19
Broken citation in Chapter 11


artur-sannikov:
Citation `network-comparison` does not exist in [Chapter 11]([network-comparison?](https://microbiome.github.io/OMA/docs/devel/pages/60_network_learning.html#ref-network-comparison)).

Relevant code [here](https://github.com/microbiome/OMA/blob/master/inst/pages/60_network_learning.qmd#L12C1-L12C282).

antagomir:
Could you make PR to fix?

artur-sannikov:
Sure, what's the reference?

antagomir:
Ah right - does @js2264 know and perhaps fix?

PR #428, opened 2024-04-21
Fix cross-references in network chapters (#425)

stefpeschel:


antagomir:
Is the solution to add these packages in rworkflows (?) - @TuomasBorman 

❯ checking package dependencies ... ERROR
  Packages suggested but not available:
    'microbiome', 'NetCoMi', 'SpiecEasi', 'SPRING', 'factoextra', 'dada2'

TuomasBorman:
Yes, that is probably the thing. However, we are able to switch our default branch from master to devel soon (this PR is pointing to devel). This issue is easier to fix when the switch is done. I or someone else can do that; then this PR should be ok without further modifications

PR #436, opened 2024-04-22
Work on aesthetics of the book

AxelDgn:
Here is a first view of the book, trying to follow Leo's instructions.
I improvised a bit with the banner if you have any ideas I'm open to suggestions. 
Let me know what you think.

OMA view 2 png
OMA view


antagomir:
It looks promising. 

For me (Ubuntu + Chromium) the logo and header are overlapping. 

Logo is also having different, lighter color than usually - it gives a less sharp impression.

R and Bioc version could be just black text (not red, like now)

antagomir:
Is more in the making?

TuomasBorman:
Looks very nice!! Is the coloring of mia logo changed or does it just look like it?

AxelDgn:
> Is more in the making?

Hello, I found a new color palette that complements the main color of the Mia logo, and I fixed some display bugs. The display on Firefox and Edge looks correct; let me know if it's good for you too. 
I just need to fix a few things before committing the changes here, but here's an overview :

image
Palette OMA

TuomasBorman:
I liked the previous one more (but if the mia logo could be with original colors)

antagomir:
YEs looks nice!

One thing I spotted and we might like to think about is the position of the references. The second TOC in right margin is a bit unnecessary but also references are handily shown on the right margin just next to the text and it would also have space for other margin notes (like footnotes). In the new version not I think. Should/could we consider that right margin can still be there but margins (both left and right) could be optionally folded out when the reader wants to have more space on the actual text? Or what should do on that..

AxelDgn:
> YEs looks nice!
> 
> One thing I spotted and we might like to think about is the position of the references. The second TOC in right margin is a bit unnecessary but also references are handily shown on the right margin just next to the text and it would also have space for other margin notes (like footnotes). In the new version not I think. Should/could we consider that right margin can still be there but margins (both left and right) could be optionally folded out when the reader wants to have more space on the actual text? Or what should

I think the left margin should stay in place, but I can check if we can add the right margin for references by adding a fold option


> I liked the previous one more (but if the mia logo could be with original colors)

I'll try to add the original logo, but this banner is a PNG image, so I was having trouble managing the CSS to make it sharp when displayed

AxelDgn:
Here is an example of what can be added in the margin : https://quarto.org/docs/blog/posts/2022-02-17-advanced-layout/

antagomir:
Great - I like to possibilty to keep both margins but it might be helpful to provide folding option for margin/s if this is not too difficult to achieve.

antagomir:
@TuomasBorman shall we merge?

TuomasBorman:
What is the layout currently? Can you paste some pictures?

TuomasBorman (approved):

TuomasBorman:
@antagomir 

I checked the layout and it is the green/yellow one. I liked the blue one more. Do you have any preferences?

TuomasBorman:
But one thing to check if we decide to go with blue one, is the banner figure. --> Is it freely available?

antagomir:
Blue is ok to me.

AxelDgn:
image

Here is the current layout, last blue banner was not free to use.

antagomir:
Is there is something still ongoing @AxelDgn

antagomir:
@TuomasBorman has approved. Can we merge?

TuomasBorman:
We can try to find free figure for that later (I think it looked very nice), but I think this is good to go now

TuomasBorman:
ONe option is to try generative AI tools

PR #439, opened 2024-04-22
Alpha diversity chapter rewritten for clarity

### LINKED ISSUE #115: Significance in boxplots
Add example on indicating significance betwen boxplots. 

There might be better options but [here is one](http://www.sthda.com/english/articles/24-ggpubr-publication-ready-plots/76-add-p-values-and-significance-levels-to-ggplots/)

Also check if (Tree)SE framework already has such visualizaiton tools readily available, and if not, consider if this should be added to *miaViz*.

antagomir:
I rewrote alpha diversity chapter and moved divergence chunk to beta diversity where it belongs.

Also linked to ggpubr pkg that provides additional significance estimation and visualization tools.

This closes #115 as well.


antagomir:
Can be merged, from my side, when the checks pass.

PR #453, opened 2024-04-24
Update in oma_packages csv

ake123:
Update in oma_packages csv
correction in section 2 of unnecessary naming
add solution to training exercise
automated method that keeps updating the oma_packages csv file #456


TuomasBorman (approved):

extract_packages.R
```diff
@@ -0,0 +1,25 @@
...
+    if (length(section_index) > 0) {
+      start_index <- section_index + 1
+      end_index <- ifelse(any(grep("^[A-Za-z]+:", lines[-(1:section_index)])),
+                          min(grep("^[A-Za-z]+:", lines[-(1:section_index)])) + section_index - 1,
+                          length(lines))
+      section_lines <- lines[start_index:end_index]
+      section_string <- paste(section_lines, collapse = " ")
+      packages <- unlist(strsplit(section_string, ","))
+      packages <- trimws(packages)
+      return(packages[packages != ""])
```
TuomasBorman: This function and also oma_package.csv file should not be in root directory (it should be as clean as possible). Instead, can you put them to directory. Maybe named oma_packages

inst/pages/04_containers.qmd
```diff
@@ -222,7 +222,7 @@ For information have a look at the [intro vignette](https://bioconductor.org/pac
...
+: **Recommended options for storing multiple data tables in microbiome studies** The _assays_ are best suited for data transformations (one-to-one match between samples and columns across the assays). The _alternative experiments_ are particularly suitable for alternative versions of the data that are of same type but may have a different number of features (e.g. taxonomic groups); this is for instance the case with taxonomic abundance tables agglomerated at different levels (e.g. genus vs. phyla) or alternative profiling technologies (e.g. amplicon sequencing vs. shallow shotgun metagenomics). For alternative experiments one-to-one match between samples (cols) is libraryd but the alternative experiment tables can have different numbers of features (rows). Finally, elements of the _MAE_ provide the most flexible way to incorporate multi-omic data tables with flexible numbers of samples and features. We recommend these conventions as the basis for methods development and application in microbiome studies. {#tbl-options}
 
 ## Demonstration data {#sec-example-data}
```
TuomasBorman, line 227: Where these changes come from? They are not related to this PR? Are they intentional?

TuomasBorman:
I guess this workflow takes max 1 minute? It is ok that it is triggered every time DESCRIPTION file is changed even though packages might still be the same

inst/pages/04_containers.qmd
ake123: Yes they are intentional and are corrected after a review sent.

TuomasBorman: Ahh, yes, good

extract_packages.R
ake123: corrected accordingly

antagomir:
OK but if it can be easily avoided to run unnecessarily then that would be even better.

PR #455, opened 2024-04-24
Cheatsheet & removing returns

ElySeraidarian:
Added link to cheatsheet & removed some return (still kept some seemed important for understanding -> can also be removed if needed)


antagomir:
73 files have changed in this PR. I expect that this is not intended for such a limited PR. Most changes don't seem to be related to the PR topic.

Can you check and possibly opening a new PR with only the intended changes?

ElySeraidarian:
Fixed the files that were added

PR #457, opened 2024-04-25
Adding link to cheatsheet in 19.1.3.2 & removed some verbose messages

### LINKED ISSUE #449: Cheatsheet
We have now phyloseq/TreeSE cheatsheet in Appendix A.5 https://microbiome.github.io/OMA/docs/devel/pages/97_extra_materials.html

1) Tidy it up (hide verbose routine rendering messages)
2) Add link to cheatsheet in section 19.1.3.2 Resources on phyloseq
3) Consider making a proper cheatsheet like this https://github.com/rOpenGov/eurostat/blob/master/inst/extras/cheatsheet/cheatsheet2019.pdf (I think there might be an R package for making cheat sheets..)

ElySeraidarian:
Fixing additionnal files that were added in last PR (#455)


inst/pages/97_extra_materials.qmd
```diff
@@ -294,7 +294,7 @@ var_explained <- e/sum(e[e>0])*100
 This section has a cheatsheet for translating common functions in phyloseq
 to TreeSE/mia with example code.
 
-```{r}
+```{r,message=FALSE, warning=FALSE, eval=FALSE}
```
antagomir: add space in "r, message" here and elsewhere

antagomir: In fact, on the same go you could move the messages, warning and eval into quarto header in chunks of this PR

Like 
#| message: false
#| warning: false
#| eval: false

ElySeraidarian: Should be better now (couldn't be able to do the way you said but pretty much the same I believe)

TuomasBorman (approved):

PR #465, opened 2024-04-27
Add Nix environment build

artur-sannikov:
It now works pretty well. With pre-cached builds, it takes a couple of minutes to build the dev environment with RStudio.


default.nix
```diff
@@ -0,0 +1,155 @@
...
+        rev = "58cc49bd118ff2b75d12b2eca006cfd48c307933";
+        sha256 = "sha256-BsQ7/nJBLadF2NQgGpHhAC10AFQaiK+HbMCIjfOH6d4=";
+      };
+      # mia dependencies (see DESCRIPTION)
+      propagatedBuildInputs = builtins.attrValues {
+        inherit (pkgs.rPackages)
+          SummarizedExperiment SingleCellExperiment TreeSummarizedExperiment
+          MultiAssayExperiment MASS ape decontam vegan BiocGenerics S4Vectors
+          IRanges Biostrings DECIPHER BiocParallel DelayedArray
+          DelayedMatrixStats scuttle scater DirichletMultinomial rlang dplyr
```
TuomasBorman: Is this same file as cachix/default.nix?

TuomasBorman: If it is possible, it is better to keep one in cachix directory since it makes root folder tidier

TuomasBorman: Also, is it necessary to build the environment every time when PR gets new commits or is it enough that it is build when devel branch gets new commit?

The book is build only when devel receives new commit

.github/workflows/oma-dev-env.yml
```diff
@@ -0,0 +1,29 @@
+name: "Update cachix cache for Orchestrating Microbiome Analysis environment"
+
+on:
+  push:
+    branches: [ devel, nix ]
```
TuomasBorman: I think we could remove "nix" branch from here (just make sure that the run works first; it is failing now)

TuomasBorman: To make the tree tidier, there should be only devel, and gh-pages branch. Other branches are temporary

TuomasBorman: --> which means that nix branch will be deleted. (if it is not required)

default.nix
artur-sannikov: These files are identical, yes. Thanks for noting that.

It's not necessary to build the environment every time actually. It's pinned to a particular commit at the moment, and that one needs to be manually updated every now and then if we want to updated packages and R version. It's also required to rebuild it if we add new packages and/or system dependencies.

I can completely remove that and set it to be manually run. Should be possible with

```yaml
on:
  workflow_dispatch:
```

artur-sannikov: The location of `default.nix` can be in `cachix`. Then, to enter the shell, users have to go to that directory, run `nix-shell`, and then exit it to have access to the entire `OMA` repo.

.github/workflows/oma-dev-env.yml
artur-sannikov: It's failing because I cannot add `CACHIX_AUTH` variable that's required to push to Cachix.

inst/pages/14_alpha_diversity.qmd
```diff
@@ -31,7 +31,50 @@ is often used to refer collectively to all these variants.
...
+  indices measure the overall community heterogeneity that considers
+  both of these aspects simultaneously. A number of ecological
+  diversity measures are available. In general, diversity increases
+  together with increasing richness and evenness. **Phylogenetic
+  diversity** (PD), [@Faith1992] is a variant that incorporates
+  information from phylogenetic relationships between species, unlike
+  most other commonly used diversity indices. The `estimateDiversity`
+  function uses a faster re-implementation of the widely used function
+  in _`picante`_ [@R_picante, @Kembel2010]. The method uses the
+  default rowTree from the `TreeSummarizedExperiment` object (`tse`).
```
TuomasBorman, line 59: Where these changes come from? They are not related to this Nix PR

artur-sannikov:
If everything is alright, I can merge it @TuomasBorman.

inst/pages/14_alpha_diversity.qmd
TuomasBorman: These things are not yet addressed

cachix/default.nix
```diff
@@ -0,0 +1,231 @@
...
+      ape
+      decontam
+      vegan
+      BiocGenerics
+      S4Vectors
+      IRanges
+      Biostrings
+      DECIPHER
+      BiocParallel
+      DelayedArray
```
TuomasBorman, line 36: Is there way to automate this thing? In the future, it might be hard to keep up with updating this, especially, if we do not have nix expert in the team. Nix heavily relies on your contribution and expertise which means that it might be good to have instructions for developers also (maybe to cachix/README.md)

artur-sannikov: What do you mean by automating?

I can write the instructions on contributing. Also there is a Matrix room with Nix R contributors that were very helpful. There is also an R package in development (https://b-rodrigues.github.io/rix/) that should simplify the transition from R to nix. I have tested it but opted to write my own `nix` file for learning purposes.

TuomasBorman: Packages are hard-coded. In optimal situation, they would be automatically fetched. You could check BiocBook, it has Docker functionality which should fetch the dependent packages also

artur-sannikov: Good, `rix` should be able to automatically fetch the packages from DESCRIPTION.

artur-sannikov:
I will open a new pull request with just the changes that are required.

Issue #330, opened 2023-08-09
Interoperability with Jl/Py


antagomir:
Show how to export TreeSE/MAE data such that it can be reused in Julia (or Python). 

The feather file format could do for this. 

Julia already has SE & MAE implemented.

This can support longer term developments.

antagomir:
@ElySeraidarian you have self-assigned this - any chance to check it through?

antagomir:
Or can we just close this issue already since the issue has been merged?

ElySeraidarian:
The section created in the OMA book explains how to export TSE/MAE in Julia and Python so I think we can close this

PR #478, opened 2024-05-13
Exporting TreeSE/MAE section

ElySeraidarian:
Issue #330 - Adding a section to explain how to export TSE/MAE data containers in order to use them in Julia or Python (using feather files). 
Not sure of the positioning of the section & the path for output file.

Note : Indicates 3 files changed but only 1 changed.


ElySeraidarian:
Build fails because can't find package feather : I don't understand where should I add it (added in DESCRIPTION & csv file)

inst/pages/04_containers.qmd
```diff
@@ -740,3 +740,39 @@ Conversion is possible between other data formats. Interested readers can refer
 * [makeTreeSEFromBiom](https://microbiome.github.io/mia/reference/makeTreeSEFromBiom.html)  
 * [importMetaPhlAn](https://microbiome.github.io/mia/reference/importMetaPhlAn.html)  
 * [importQZA](https://microbiome.github.io/mia/reference/importQIIME2.html)
+
+## Exporting data container {#sec-exporting-data-container}
+
+### Export TreeSummarizedExperiment
+
+Exporting a `TreeSE` data container can be done using *feather* package. TSE object has to be converted into a dataframe (`DataFrame`). The output file is a .feather file, which can be imported in other languages such as Julia or Python. For information, have a look at the [FeatherFile](https://juliapackages.com/p/featherfiles) Julia package and [feather-format](https://pypi.org/project/feather-format/) Python library.
```
TuomasBorman: Does also data.frame object work? DataFrame and data.frame are different. DF inherits data.frame. One option is also basic csv file (you could add that).

inst/pages/04_containers.qmd
```diff
@@ -740,3 +740,39 @@ Conversion is possible between other data formats. Interested readers can refer
...
+
+```{r, eval=FALSE}
+# Import TSE
+library(mia)
+data(GlobalPatterns, package="mia")
+tse <- GlobalPatterns
+# Convert into a DataFrame
+tse <- transformAssay(tse, MARGIN = "samples", method="relabundance")
+molten_tse <- mia::meltAssay(tse,
```
TuomasBorman: remove mia::

inst/pages/04_containers.qmd
```diff
@@ -740,3 +740,39 @@ Conversion is possible between other data formats. Interested readers can refer
...
+Exporting a `MultiAssayExperiment` data container can also be done using *feather* package.
+
+```{r, eval=FALSE}
+# Convert into a DataFrame
+molten_mae <- longFormat(mae)
+molten_mae <- data.frame(molten_mae)
+path <- "dirname(rstudioapi::getSourceEditorContext()$path)/../../../mae.feather"
+write_feather(molten_mae,path)
+```
+
```
TuomasBorman, line 797: Remove extra line (only one line)

inst/pages/04_containers.qmd
```diff
@@ -740,3 +740,39 @@ Conversion is possible between other data formats. Interested readers can refer
...
+### Export MultiAssayExperiment
+
+Exporting a `MultiAssayExperiment` data container can also be done using *feather* package.
+
+```{r, eval=FALSE}
+# Convert into a DataFrame
+molten_mae <- longFormat(mae)
+molten_mae <- data.frame(molten_mae)
+path <- "dirname(rstudioapi::getSourceEditorContext()$path)/../../../mae.feather"
```
TuomasBorman: You could add just arbitrary path. I added eval=FALSE, so that nothing is really saved

TuomasBorman: like "path/to/mae.feather"

antagomir:
up

ElySeraidarian:
I have made fixes after review but build fails because package not found. However works locally when building the book (imports the package).

antagomir:
@ake123 might know how to solve?

ElySeraidarian:
One way to solve it would be just removing the package from build as we use eval=FALSE it will not be an issue. However, if the user wants to use the code, he will have to install it manually

antagomir:
Check with @ake123 if there is a real solution, and resolve the remaining comments.

ake123:
I forked his repo and tested it and seems to be working fine

antagomir:
> I forked his repo and tested it and seems to be working fine

Yes but how about the failing builds?

ake123:
It's strange because I tested it here (https://github.com/Akejeb/OMA_ely/actions). seems to be fine for me.

antagomir:
How shall we proceed in order to get this solved - any suggestion?

TuomasBorman:
I think it is not wise to save anything, so we can just put eval=FALSE and show how it can be done. There is probably no informative message about saving part so nothing is lost

Then we can remove the package from dependencies

TuomasBorman:
This issue of package that cannot be found is caused because GHA saves cache. If the version is not bumped, GHA expects that the same cache can be used and no new packages are installed (or that is what I think)

TuomasBorman:
Ohh, eval=FALSE already. SO just remove the package from DESCRIPTION. That is also good in that way that we have less dependencies

ElySeraidarian:
Here we are

antagomir:
merge?

PR #489, opened 2024-05-21
Add Dark mode

0xMuluh:


TuomasBorman:
It seems that the darkmode does not work?

TuomasBorman:
The button is there but it does not do anything

0xMuluh:
I have added a fix for the dark mode.

TuomasBorman (approved):
Wow! Thanks

PR #502, opened 2024-05-29
Improve biclustering plot readability

artur-sannikov:
Now the plots are readable. However, especially for the last plot there should be a consideration to reduce the number of features because now the plot is large and column and row names are tiny.


antagomir (approved):
Very nice, thanks!

PR #534, opened 2024-06-24
Update 11_taxonomic_information.qmd

jkc9886:
Grammatical and Punctuation corrections.

### DIFF
--- a/inst/pages/11_taxonomic_information.qmd
+++ b/inst/pages/11_taxonomic_information.qmd
@@ -24,10 +24,10 @@ and assumes certain key aspects:
 * The columns containing the taxonomic information must be named `domain`,
 `kingdom`, `phylum`, `class`, `order`, `family`, `genus`, `species` or with
 a capital first letter.
-* the columns must be given in the order shown above
+* The columns must be given in the order shown above.
 * column can be omitted, but the order must remain
 
-In this chapter, we will refer to co-abundant groups as CAGs, which are
+In this chapter, we will refer to the co-abundant groups as CAGs, which are
 clusters of taxa that co-vary across samples.
 
 ## Assigning taxonomic information
@@ -120,10 +120,10 @@ head(getUnique(tse, rank = "Phylum"))
 
 ### Generate a hierarchy tree on the fly {#sec-fly-tree}
 
-A hierarchy tree is not an actual phylogenetic tree.
+A hierarchy tree is not an actual phylogene tic tree.
 A hierarchy tree shows mapping between the taxonomic levels in taxonomic rank table (included in rowData), rather than the detailed phylogenetic relations included in a phylogenetic tree
 
-For instance, the TreeSummarizedExperiment dataset Tengeler2020 contains an actual phylogenetic tree.
+For instance, the TreeSummarizedExperiment dataset, Tengeler2020, contains an actual phylogenetic tree.
 
 ```{r}
 data(Tengeler2020)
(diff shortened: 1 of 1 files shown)


inst/pages/11_taxonomic_information.qmd
```diff
@@ -120,10 +120,10 @@ head(getUnique(tse, rank = "Phylum"))
 
 ### Generate a hierarchy tree on the fly {#sec-fly-tree}
 
-A hierarchy tree is not an actual phylogenetic tree.
+A hierarchy tree is not an actual phylogene tic tree.
```
antagomir: space typo?

inst/pages/11_taxonomic_information.qmd
```diff
@@ -336,22 +336,22 @@ references.
    in microbiome-based classification performance [@Giliberti2022,
    Karwowska2024].
 
- * 'z' Z transformation scales data to zero mean and unit variance;
-   this is used to bring features (or samples) to more comparable
+ * 'z' Z transformation scales data to zero mean and unit variance.
```
antagomir: This should be updated to "scale" (we renamed it recently).

antagomir (approved):

inst/pages/11_taxonomic_information.qmd
antagomir: Sorry, in fact it is "standardize" (and not "scale") I think. In the latest mia version.

inst/pages/11_taxonomic_information.qmd
```diff
@@ -329,29 +329,29 @@ references.
...
    widely used in ecological research. It has shown good performance
    in microbiome-based classification performance [@Giliberti2022,
    Karwowska2024].
 
- * 'z' Z transformation scales data to zero mean and unit variance;
-   this is used to bring features (or samples) to more comparable
-   levels in terms of mean and scale of the values. This can enhance
+ * 'standardize' Standardize transformation scales data to zero mean and 
+   unit variance. This is used to bring features (or samples) to more 
+   comparable levels in terms of mean and scale of the values. This can enhance
```
TuomasBorman: Add "(or Z-transformation)"

antagomir: I would prefer not referring to z-transformation any more since it has different common meaning.

TuomasBorman: Ok, that is fine. I guess Z-score is relatively common though. The description already states clearly what the transformation does, but for someone Z-score might be more familiar term as "standardization" can also mean different things

antagomir: Can you point us to at least 1-3 references where other than us use z transformation in our meaning in microbiome analyses?

TuomasBorman: With a quick search (I have not done that earlier; my point was that it might be good to list common synonyms from basic statistics even though they are not commonly used in microbiome research.)

- https://microbiomejournal.biomedcentral.com/articles/10.1186/s40168-016-0222-x/figures/2
- https://www.researchgate.net/publication/338713495_Diversity_of_Bacterial_Community_in_the_Oxygen_Minimum_Zones_of_Arabian_Sea_and_Bay_of_Bengal_as_Deduced_by_Illumina_Sequencing/figures?lo=1
. https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8250139/

antagomir: This also discusses "z-score" in the same meaning: https://en.wikipedia.org/wiki/Standard_score

The "z" has multiple meanings like "z score", "z value" "z factor" etc.
https://en.wikipedia.org/wiki/Z-score_(disambiguation)

Seems fine to refer to this also as z-score.

jkc9886:
This PR is ready to be merged.

antagomir (approved):
Good to me!

antagomir:
This seems OK - to merge, @TuomasBorman ?

TuomasBorman (approved):

PR #561, opened 2024-07-19
Header update

TuomasBorman:
I liked the style in this PR https://github.com/microbiome/OMA/pull/436.

This PR adds illustrative figure to header. The figure can be easily removed since it is only in background of the header (blue color would still be there). I still have to find free images and add citation. With these modification, the page looks like this: 
image


TuomasBorman:
Also I changed the header color from random blue  to Bioconductor blue

TuomasBorman:
I found the source of the figure: https://www.istockphoto.com/fi/valokuva/mikroskooppisen-sinisen-bakteerin-tausta-gm628978952-111796005

The license is ok for our use, I think

antagomir:
Seems good! 

Title options:
- Orchestrating microbiome analysis
- Orchestrating microbiome data science
- Orchestrating microbiome analysis with Bioconductor
- Orchestrating microbiome data science with Bioconductor

Data science might be useful to have in the title (analysis can also refer to other than computational methods). 

Bioconductor in title would be informative and in line with OSCA (https://bioconductor.org/books/release/OSCA/) and [OSTA](https://bioc2021.bioconductor.org/posts/paper74/)

Issue #352, opened 2023-10-04
Structure of OMA


TuomasBorman:
Consider structure of OMA. How much general information there should be about microbiome data science (not directly related to miaverse)? Should we structure the book differently? Currently it seems that the material is just put somewhere.

It could be beneficial if we have a book that gathers all relevant information and directs the reader to correct source material. 

The structure could be:

1. General information on miaverse
   - What is miaverse
   - How miaverse fits to microbiome data science workflow
2. General information on microbiome research
   - What and why?
   - What are the samples?
   - How to design a study.
3. How to analyze / measure DNA from samples
   - What sequencing approaches are available.
   -  How to get abundance tables. (HuMANn workflow, DADA2)
4. Downstream analytics
   - Basic topics
   - Advanced topics
5. Exercises
6. Other material

TuomasBorman:
Yes, the purpose is not to copy those resources but rather create a platform from where to find relevant resources in microbiome data science. The focus is still on downstream analytics but I think background information is important for several reasons:

1. Supports the user. There are lots of information and it is hard to find the relevant info.  (That is also my own experience from spring when I ran HuMANn worklfow). Even though there are good tutorials, how to find the "best" tools? OMA could be the main resource from where to find information.
2.  If people start to use OMA as a main resource, we could get more users.
3. There are lots of information that for example our group has but they are not really written anywhere

- If one wants to design a study --> one can find information (or rather good papers) from beginning to end
- If data scientist without microbiome knowledge wants to start analyze microbiome --> OMA could have background information
- If wet lab scientist want to analyse data --> OMA has information how to get abundance tables and then use miaverse

I believe there is no that kind of resource that gathers the information which is novel and valuable in that sense, I believe. I agree that background information must not be too long.

TuomasBorman:
Any thoughts? @microsud

antagomir:
Exactly, this is my concern. These topics are extremely vast, diverse, _and_ fast-evolving. We would then need at least a clear and plausible plan who will write those materials and how them could be kept up-to-date in addition to the code base, which may already in itself prove to be challenging. At the minimum, I would try to stabilize and organize better the current contents before initiating substantial new extensions. 

I do agree that the book could give pointers to good practices. Having another look at other tutorials (QIIME2, Mothur, etc.) can also be helpful for inspiration.

TuomasBorman:
Good points. I agree that let's now focus on finishing this miaverse workflow part.

[...]

PR #562, opened 2024-07-20
Restructure the book

TuomasBorman:
Hello,

https://github.com/microbiome/OMA/issues/352

here is my suggestion for the structure of OMA. You can see the book by extracting the docs zip file, and by opening index.html. Below is the file structure of the book to give some idea of the structure.

[docs.zip](https://github.com/user-attachments/files/16321444/docs.zip)

- part: "Introduction"
      chapters:
        - pages/intro.qmd
        - pages/miaverse.qmd
- part: "Data containers & Importing"
  chapters:
	- pages/containers.qmd
	- pages/import.qmd
	- pages/convert.qmd
- part: "Data Manipulation"
  chapters:
	- pages/wrangling.qmd
	- pages/transformation.qmd
	- pages/taxonomy.qmd
	- pages/agglomeration.qmd
- part: "Exploration & QC"
  chapters:
	- pages/quality_control.qmd
	- pages/composition.qmd
- part: "Diversity & Similarity"
  chapters:
	- pages/alpha_diversity.qmd
	- pages/beta_diversity.qmd
	- pages/clustering.qmd
- part: "Association"
  chapters:
	- pages/differential_abundance.qmd
	- pages/correlation.qmd
	- pages/msea.qmd
	- pages/mmuphin_meta_analysis.qmd
- part: "Networks"
  chapters:
	- pages/network_learning.qmd
	- pages/network_comparison.qmd
- part: "Multiomics"
  chapters:
	- pages/cross_correlation.qmd
	- pages/multiassay_ordination.qmd
	- pages/integrated_learner.qmd
- part: "Machine Learning & Statistical Modeling"
  chapters:
	- pages/machine_learning.qmd
	- pages/statistical_modeling.qmd
- part: "Workflows"
  chapters:
	- pages/introductory_workflow.qmd
	- pages/introductory_workflow_french_version.qmd
	- pages/introductory_workflow_dutch_version.qmd
- part: "Training"
  chapters:
	- pages/training.qmd
	- pages/exercises.qmd
- part: "Support & Resources"
  chapters:
	- pages/support.qmd
	- pages/resources.qmd
	- pages/acknowledgments.qmd
appendices:
- pages/extra_material.qmd
- pages/visualization.qmd
- pages/session_info.qmd

The idea is that the structure is as intuitive and clear as possible allowing us to expand it in the future as needed. Also important point is that the reader should be able to find easily the information needed (there is search option, but that does not help too much if the structure is not clear).

Moreover, I found sections that were repetitive. My plan was to modify the book so that no information was lost. The repetitions made the structure unclear. Now there is linkages between related sections.

Couple points that I made. These things should be fixed or taken into account otherwise:

1. I used callout blocks (note, tip, important and warning). They look quite nice so we could use them more.

2. We could focus more on interpretations as they are something that readers want. They are the hard part which is why we should focus more on them.

3. We use both styles `library(mia)` and `library("mia")`. We should have only one style. I prefer `library("mia")`.

4. In the text, we should harmonize the style of package and function names. Currently, we use several styles. We could have: \`package\`, \`function()\` and \`assay\` slot.

5. Exercises are outdated. Especially RDA exercises as they use vegan directly even though we have simple addRDA() wrapper available.

For the future maintenance, I have also couple points:

1.

We should not exceed 80 characters in line. In the text, it is not that strict but in the code it makes harder to see the code if we have long lines (code does not fit into page correctly). For example, there were this kind of lines quite many

```
long_variable_name <- long_function_name(parameter = "parameter".
                                         parameter2 = "parameter2")
```

Instead I would prefer this kind of lines as it make easier to see the parameters

```
long_variable_name <- long_function_name(
    parameter = "parameter".
    parameter2 = "parameter2"
	)
```

2.

All the chunks should have labels. That is not urgent, but we should take this into account when the code is modified. All modified chunks should have label in the future.

Please see the structure and give feedback. It is much easier to change now than in the future (bigger changes are not probably even possible anymore when the book is established).

-Tuomas


antagomir:
To me this looks good, no further suggestions at this point.

I like more the style library(mia) than library("mia") as I like it shorter. But can be either one.

Maybe need to open issues about outdated examples and other things that are not feasible to solve as part of the structure update.

The last three items (extra material, visualization, session) could be under appendix chapter, or this is already the idea?

TuomasBorman:
Updated README's contribution instructions and made issues on points that need to be resolved.

library(mia) is fine. My justification was that then `library("mia")` is consistent with `install.packages("mia")`.

Last three sections are already under appendix. My mistake with indentation in the comment above.

Merging when checks are ok.

antagomir:
One more thing to do is to make a quick check on [OSCA](https://bioconductor.org/books/release/OSCA/) and its sub-books for possible inspiration and ideas that we may have overlooked.

TuomasBorman:
I checked the OSCA (and some of the carpentries online book) already for inspiration. Might be good to have other's thoughts also. Moreover, there is room to improve the text. If you have time to do that it would be nice.

PR #583, opened 2024-07-25
Add gtools to DESCRIPTION

artur-sannikov:
When running R session in container, it was saying that `gtools` in not in DESCRIPTION, but it appears to be inside the container according to `SessionInfo()`.


antagomir (approved):

antagomir:
Thanks

antagomir:
If there is an issue related to this, can you close it (as a standard practice) @artur-sannikov

Issue #385, opened 2024-01-10
Improve Docker container


artur-sannikov:
Right now[ Docker container](https://github.com/js2264/OMA/pkgs/container/OMA) is very basic and cannot be used.

I managed to run in with [Podman](https://podman.io/) 4.8.3 :

```bash
podman pull https://github.com/js2264/OMA/pkgs/container/OMA
podman run --rm -p 8787:8787 <container_id>
```
It's then accessible on localhost:8787. Username `root` and password is given in shell.

There are no accessible files in RStudio web-interface. Inside the container there is a `OMA_0.98.16.tar.gz` archive.

[BiocBook](https://bioconductor.org/packages/devel/bioc/vignettes/BiocBook/inst/doc/BiocBook.html) supports containerization, so it a Docker image might be nice to have because it can help avoid all dependency issues on different platforms and their versions people run.

TuomasBorman:
Thanks! We will take this into account

artur-sannikov:
You now need to run `podman pull ghcr.io/js2264/oma:latest`.

antagomir:
@artur-sannikov is this ready?

artur-sannikov:
@antagomir, currently the most up-to-date container is on `ghcr.io/microbiome/oma`, but it's private and requires logging in with `docker login ghcr.io`. I just tried it on a fresh install of Fedora server and it fails at 

```
[23/37] pages/multiassay_ordination.qmd


processing file: multiassay_ordination.qmd
  |..........................................               |  74% [mofa6]     Error in `py_get_attr()`:
! AttributeError: module 'mofapy2' has no attribute 'run.entry_point'
Run `reticulate::py_last_error()` for details.
Backtrace:
 1. MOFA2::run_mofa(model, use_basilisk = TRUE)
 2. basilisk::basiliskRun(...)
 3. MOFA2 (local) fun(...)
 4. MOFA2:::.run_mofa_reticulate(object, outfile, save_data)
 6. reticulate:::`$.python.builtin.module`(mofa, "run.entry_point")
 7. reticulate::py_get_attr(x, name, FALSE)



Quitting from lines 126-136 [mofa6] (multiassay_ordination.qmd)
                                                                                                       
Execution halted
Error in run_serve_daemon("preview", file, NULL, args, render, port, host,  : 
  Error starting quarto
```

Image `ghcr.io/js2264/oma:latest` is outdated.

TuomasBorman:
[...]

PR #585, opened 2024-07-26
Install Python and miniconda for MOFA+

artur-sannikov:
Should fix the error in https://github.com/microbiome/OMA/issues/385#issuecomment-2244487594.


antagomir (approved):

antagomir:
Seems good to me.

PR #589, opened 2024-07-29
removing polyfill 

ake123:
removing polyfill


TuomasBorman:
Note that only devel branch's GHA should push to gh-pages. This is temporal fix that does not fix the real problem. These fixes are overwritten by next push to devel (because GHA then pushes new html files to gh-pages).

ake123:
Yes true and trying to find the root cause. The thing is the polyfill io script is replaced by other but it still exists.
The other thing is I am suspecting it is coming from some visualization package such as plotly

ake123:
I actually got a permanent solution by disabling mathjax on the quarto yaml

PR #592, opened 2024-08-05
Trial : creating DAA_interactivetable

jkc9886:
1. Trying to add a hidden chapter that can be accessed via a direct url and contains data and code to an interactive table comparing DAA methods.

2. This chapter should be accessed at https://github.com/microbiome/OMA/tree/devel/inst/pages/DAA_interactivetable after the PR is merged. Then I can add the link to an interactive table trial in the DAA chapter.

3. I have not added the table's code yet because this PR is just a trial attempt to see if the method works.


antagomir:
Which DAA methods are included, and what measures you plan to compare?

jkc9886:
I plan on comparing the results of different Differential Abundance Analysis (DAA) methods, such as ALDEx2, ANCOM-BC, and MaAsLin2. By selecting different methods, we try to visualize and compare the differentially abundant taxa identified by each, along with their p-values and effect sizes.

antagomir:
Ok. Those 3 methods might be sufficient.

antagomir:
Status of this one?

antagomir:
@jkc9886 above it is said this is a draft - is it still ongoing, or ready?

TuomasBorman:
There has been no activity, and the file seems to be rather empty. I suggest that we close this PR for now. We can open this again, when there are something ready to review.

This might be something to align with [benchdamic package](https://www.bioconductor.org/packages/release/bioc/html/benchdamic.html)

I will keep the branch still available.

PR #597, opened 2024-08-09
Modify the NMF example to use new getNMF wrapper

theotimeee:
After PR https://github.com/microbiome/mia/pull/616, this PR modifies the NMF example in OMA.

The output of new wrapper `getNMF` is briefly explained, and the wrapper is used to calculate the results shown in the example.


antagomir (approved):
I think this can be merged. Loadings could also be in rowData but I think this we discussed already.

antagomir (approved):

inst/pages/clustering.qmd
```diff
@@ -323,27 +323,29 @@ continuous sample memberships across all community types. NMF algorithm
...
+# Extract NMF feature loadings
+loadings <- attr(res, "loadings")
 
 # Add NMF loadings in the metadata
-metadata(altExp(tse, "prevalent"))$NMF_loadings <- H
+metadata(altExp(tse, "prevalent"))$NMF_loadings <- loadings
 
 # Obtain the relative abundance of ES in each sample
-wrel <- t(apply(W, 1, function (x) {x/sum(x)}))
+wrel <- t(apply(res, 1, function (x) {x/sum(x)}))
```
TuomasBorman: SOrry for this long delay. I have been extremely busy

Why not to use addNMF as it stores the results to TreeSE?

TuomasBorman:
> I think this can be merged. Loadings could also be in rowData but I think this we discussed already.

This same thing applies to other functions which returns ordination and loadings. It could be simpler to add loadings to rowData() (easier to find them from there), however, I do not know how feasible the solutions is if there are many components (it might be ok). However, if the rest of results are stored in metadata/reducedDim, it might be confusing

antagomir:
> > I think this can be merged. Loadings could also be in rowData but I think this we discussed already.
> 
> This same thing applies to other functions which returns ordination and loadings. It could be simpler to add loadings to rowData() (easier to find them from there), however, I do not know how feasible the solutions is if there are many components (it might be ok). However, if the rest of results are stored in metadata/reducedDim, it might be confusing

Most improtant that we are consistent. One problem with rowData is that the rowData columns should then be recognized by their name (or does the structure allow having multiple separate rowData elements.. or can rowData include lists..)

TuomasBorman:
> > > I think this can be merged. Loadings could also be in rowData but I think this we discussed already.
> > 
> > 
> > This same thing applies to other functions which returns ordination and loadings. It could be simpler to add loadings to rowData() (easier to find them from there), however, I do not know how feasible the solutions is if there are many components (it might be ok). However, if the rest of results are stored in metadata/reducedDim, it might be confusing
> 
> Most improtant that we are consistent. One problem with rowData is that the rowData columns should then be recognized by their name (or does the structure allow having multiple separate rowData elements.. or can rowData include lists..)

rowData can have any kind of object (I think) that has as many elements or rows as TreeSE has rows.


```
data("GlobalPatterns")
tse <- GlobalPatterns

# Does work
rowData(tse)[["asd"]] <- assay(tse)
rowData(tse)[["asd2"]] <- as.list(rownames(tse))

# Does not work
rowData(tse)[["asd"]] <- assay(tse) |> t()
rowData(tse)[["asd2"]] <- as.list(colnames(tse))

```

TuomasBorman (approved):

Issue #614, opened 2024-09-04
Improve examples


TuomasBorman:
Hello,

some of the examples in OMA are too complex. The examples should be clear enough without unnecessary complexity because it might a) prevent readers to get interested on the framework, and b) make harder for readers to troubleshoot possible problems in their code

-  [x] 1. Cite to Bioconductor if the package can be found from there
https://microbiome.github.io/OMA/docs/devel/pages/miaverse.html#other-relevant-packages

Check these examples

-  [x]  2. Check if importBIOM has option for polishing the names.

The rowData() polishing step is little bit off-scope. I think it would be better to remove it.

https://microbiome.github.io/OMA/docs/devel/pages/import.html#standard-data-formats

-  [x]  3. From all examples, replace %>% with |>

Unifies the code.

-  [x] 4. Paragraph unclear

https://microbiome.github.io/OMA/docs/devel/pages/wrangling.html#subset-by-sample-column-wise

"This chapter covers various basic techniques for preparing data for subsequent analysis. These include subsetting, splitting, and modifying data, as well as converting data to a data.frame through melting and merging multiple SummarizedExperiment objects. See Chapter 3 for basics of TreeSE."

-->

The list could be improved. For instance, it seems that converting to df requires merging multiple SEs which is not the case.

-  [x] 5. Remove since other examples do not have this kind of package listing

https://microbiome.github.io/OMA/docs/devel/pages/wrangling.html#sec-treese_subsetting

Load:

    mia
    dplyr
    knitr
    data GlobalPatterns
	

-  [x] 6. "Note" block should be in the end of the example.

Remve the paragraph starting with "As a sanity check,..." 

https://microbiome.github.io/OMA/docs/devel/pages/wrangling.html#subset-by-sample-column-wise

-  [x] 7.

Paragraph "Several criteria can be used to subset by sample:" does not make sense (at least like this.) It could be removed. The subsetting is dependant on research question and listing these like this does not seem right.

-  [x] 8. Combine 6.1.2 and 6.1.3

These are actually referring to same example.

https://microbiome.github.io/OMA/docs/devel/pages/wrangling.html#subset-by-feature-row-wise

-  [x] 9. Add link to agglomeration chapter

In agglomerateByRanks, add link to agglomerations chapter

https://microbiome.github.io/OMA/docs/devel/pages/wrangling.html#splitting

-  [ ] 10. How to add data to assay

Create an example that adds new sample

[...]

PR #615, opened 2024-09-05
Improve examples

0xMuluh:
link #614

### DIFF
--- a/inst/pages/extra_material.qmd
+++ b/inst/pages/extra_material.qmd
@@ -56,10 +56,10 @@ rowData commands.
 ```{r}
 phy_taxtable <-
     tax_table(phy)  |> # Access the phyloseq taxonomic name table
-    data.frame # Make into a data frame
+    data.frame() # Make into a data frame
 
 tse_taxtable <- rowData(tse) |> # Same for tse
-    data.frame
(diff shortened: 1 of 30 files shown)


inst/pages/extra_material.qmd
```diff
@@ -55,10 +55,10 @@ rowData commands.
 
 ```{r}
 phy_taxtable <-
-    tax_table(phy)  %>% # Access the phyloseq taxonomic name table
+    tax_table(phy)  |> # Access the phyloseq taxonomic name table
     data.frame # Make into a data frame
 
-tse_taxtable <- rowData(tse) %>% # Same for tse
+tse_taxtable <- rowData(tse) |> # Same for tse
     data.frame
```
TuomasBorman: Always run the code. This does not work.

TuomasBorman (approved):

TuomasBorman:
This is good to go. Thanks!!!

Can you in different PR check those exercises. It seems that some of them leads to an error.

1. Remove eval = FALSE from individual chunks
2. Add global chunk option in the beginning of the file --> eval = TRUE
3. Check that the solution a) run and b) are correct

Issue #580, opened 2024-07-23
Combine all languages into one workflow page


RiboRings:
Hi! Probably not very urgent now, but if the [workflow](https://microbiome.github.io/OMA/docs/devel/pages/introductory_workflow.html) will be translated to other languages, [Quarto profiles](https://quarto.org/docs/projects/profiles.html) could be used to combine all languages into one document and show only one selected language at a time. This is somewhat explained in [this tutorial](https://quarto-dev.marioangst.com/en/blog/posts/multi-language-quarto/), but I don't know if it's feasible with the OMA book.

TuomasBorman:
This might be nice to have before September if possible. Can @ake123 have a look?

ake123:
Yes sure that's a great idea. I will check if it's feasible.

TuomasBorman:
I think we could consider moving the workflows to here: https://github.com/microbiome/miaDemo

PR #631, opened 2024-10-15
use profiles for language

0xMuluh:
https://github.com/microbiome/OMA/issues/580

### DIFF
--- a/inst/_quarto.yml
+++ b/inst/_quarto.yml
@@ -7,3 +7,23 @@ metadata-files:
   - assets/_website.yml
   - assets/_format.yml
   - assets/_knitr.yml
+profiles:
+  english:
+    book:
+      language: en
+    project:
+      input: "introductory_workflow.qmd" 
+
+  german:
+    book:
+      language: de
+    project:
+      input: "introductory_workflow_dutch_version.qmd" 
+
+  french:
+    book:
+      language: fr
+    project:
+      input: "introductory_workflow_french_version.qmd" 
+      
+default-profile: english


0xMuluh:
@TuomasBorman

TuomasBorman:
Did you render the book? I rendered it but it seems that these modifications do not have any effect?

0xMuluh:
> Did you render the book? I rendered it but it seems that these modifications do not have any effect?

Yes, for now they don't have any direct effect but would only be useful if there is a long term ambition to translate OMA to other languages where for example we could link qmd sections to specific profiles: `{when-profile="english"}`. However, I could add a language selector in the navbar for a start if this is something that is feasible. Without which we can probably ignore for now.

TuomasBorman:
Translating the book would be nice but it is not realistic (at least currently). Translating it to different language would require lots of work and cause maintenance issues. OMA book is updating continuously and we do not have time or expertise to keep the translated versions up-to-date. In fact, the tutorial workflows are not up-to-date either.

If this current PR does not modify the book, I think we can close the PR and modify the book when needed.

Would it be possible to add this kind of option only for the tutorial workflow (so that it does not affect the whole book)? That might be feasible.

Moreover, we have to decide at some point the correct place for these workflows. The optimal place might be in OMA, but  there could be separate book for them. It might be problematic if we have the same workflow multiple times with different languages in OMA.

0xMuluh:
> Translating the book would be nice but it is not realistic (at least currently). Translating it to different language would require lots of work and cause maintenance issues. OMA book is updating continuously and we do not have time or expertise to keep the translated versions up-to-date. In fact, the tutorial workflows are not up-to-date either.
> 
Right.

> If this current PR does not modify the book, I think we can close the PR and modify the book when needed.
> 

I agree

> Would it be possible to add this kind of option only for the tutorial workflow (so that it does not affect the whole book)? That might be feasible.
> 
We currently have a hard-coded language selector in the tutorial workflow that helps in switching the languages. Since multi-language is not needed immediately the option is not really necessary. Though it could be a start in managing updates across different languages in a single sweep 

> Moreover, we have to decide at some point the correct place for these workflows. The optimal place might be in OMA, but there could be separate book for them. It might be problematic if we have the same workflow multiple times with different languages in OMA.

right. it seems the workflows could be rendered per profile basis. and as you have rightly put, having separate book could be a good thing but the downside would be the complexity of updating the books simultaneously. alternatively different language profiles could be stored in subdirectories while the primary language used the main index. 
Another thing could also be the comments in the code snippets across the languages. Are these of interest in the translation as well?

TuomasBorman:
Good points. I think we should have only translations in tutorial workflow (if any). This should not take our time too much.

Comments are not necessary if the translation is otherwise ok. (But currently the code chunks also differ in different languages of tutorial workflows...)

--> I suggest that we close this PR now and solve this issue when OMA is more ready. (These workflows can also be moved to extra material.) Anyways, we are not going to translate the whole book, so these changes in this PR are not required

0xMuluh:
sounds good. closing pr now.

Issue #635, opened 2024-11-11
Polish ML chapter


TuomasBorman:
Improve ML chapter: https://microbiome.github.io/OMA/docs/devel/pages/machine_learning.html#supervised-machine-learning

Use plotLoadings in xgBoost loadings, if possible. Remove unecessary printing, for instance:
```

p
##  NULL
```

PR #636, opened 2024-11-13
suppress non informative output in ml chapter

0xMuluh:
ping #635

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: OMA
 Title: Orchestrating Microbiome Analysis with Bioconductor
-Version: 0.98.29
+Version: 0.98.30
 Date: 2024-10-04
 Authors@R:
     c(person("Leo", "Lahti", role = c("aut"),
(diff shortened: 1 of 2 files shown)


TuomasBorman (approved):

TuomasBorman:
We use ROC plots that are not ggplot. Can you switch to ggplot? Use for instance plotROC package. Remember to update DSCRIPTION file

TuomasBorman:
And try this

```
library(xgboost)
library(miaViz)

# Get feature importance
df <- xgb.importance(model = model$finalModel) |> as.data.frame()
rownames(df) <- df[["Feature"]]
df <- df[, c("Gain"), drop = FALSE]
df <- as.matrix(df)

# Create a plot
p <- plotLoadings(df, ncomponents = 1, n = 20, show.color = FALSE)
p
```

TuomasBorman:
I removed unused ROCR from DESCRIPTION. Moreover, if you add packages, you have to bump the version. Otherwise the environment is not updated

PR #671, opened 2025-02-26
expand supervised ML chapter

Benjamin-Valderrama:
Added new content, code and references to the chapter. Removed a library that is not used anymore from DESCRIPTION.


inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
-insights into how well the model might perform on new data.
+Common tasks for supervised machine learning includes classification (e.g.,
+predict categorical variables) and regression (e.g., predicting continuous
+variables). The focus of this chapter are classification tasks, but note that
+some ML models discussed below can also be used for regression.
 
+::: {.callout-note}
+## Note: ML in multi-omics data analysis
+ML applications for the integration of multi-omic datasets is covered in
+[@sec-cross-correlation], [@sec-multiassay_ordination] and
```
TuomasBorman: @sec-cross-correlation is just about calculating correlations, so it can be removed from here

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+    samples, 
+    dataType = "relative_abundance",
+    counts = TRUE, 
+    rownames = "short"
+    )
 
 table(tse[["disease"]])
+##
+## healthy     T2D 
+##     193     170
```
TuomasBorman: This is printed so the comment can be removed

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+#| output: FALSE
+
+library(mia)
+
+# Preprocessing step 1 - 
+# Keep taxa present in more than 10% of samples
+tse_prev <- subsetByPrevalent(tse, 
+                              assay.type = "relative_abundance",
+                              prevalence = 10/100)
+
```
TuomasBorman: To harmonize the style, use:

```
tse_prev <- subsetByPrevalent(
    tse, 
    assay.type = "relative_abundance",
    prevalence = 10/100
)
```

with 4 spaces indentation

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+# calculate all available alpha diversity measures
+vars_before <- colnames(colData(tse))
+tse_prev <- addAlpha(tse_prev, assay.type = "relative_abundance")
 
-# We calculate all available alpha diversity measures
-variables_before <- colnames(colData(tse))
-tse <- addAlpha(tse, assay.type = "relative_abundance")
 # By comparing variables, we get which indices were calculated
-index <- colnames(colData(tse))[ !colnames(colData(tse)) %in% variables_before ]
+index <- colnames(colData(tse_prev))[!colnames(colData(tse_prev)) %in% vars_before]
```
TuomasBorman: This could be done simpler way by retrieving all column names that have "diversity", "evenness" or other suffix

inst/pages/machine_learning.qmd
TuomasBorman: Check also other chunks

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+cor_df <- data.frame(row = rownames(m)[as.vector(row(m))],
+                     col = colnames(m)[as.vector(col(m))],
+                     cor = c(m)) |>
+        filter(row > col & abs(cor) == 1)
+##
+##                      row                    col  cor
+## simpson_lambda_dominance gini_simpson_diversity   -1
+##       relative_dominance          dbp_dominance    1
+##        observed_richness         chao1_richness    1
+
```
TuomasBorman: mikropml does have a function that can be used to do this --> use it instead

TuomasBorman: https://www.schlosslab.org/mikropml/reference/preprocess_data.html

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+After the preprocessing is done, different supervised ML models can be used to
+classify samples. In this section, two of the most used supervised ML algorithms
+are discussed: 
+
+* Random Forest
+* XGBoost
+
+Other resources are suggested for an introduction to the mathematical
+underpinnings of these -and other- models [@james2013_ISLR].
```
TuomasBorman: emdash can be done with `&mdash;`

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
-)
-
-# Get predictions
-res <- model[["pred"]]
-# Print result
-confusionMatrix(data = res[["pred"]], reference = res[["obs"]])
-```
+Although this chapter focuses in the area under the receiver operator 
+characteristic curve (AUC), it is often a good idea to look at different 
+metrics to accurately assess the model's performance.
```
TuomasBorman: Also if there are more than  2 classes.

It might be good to add one sentence about accuracy and how it can be misleading if the class balance is bad

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
-# Prepare data for ROC
-roc_data <- data.frame(
-    observed_class = as.numeric(res[["obs"]] == "healthy"),
-    predicted_probability = res[["healthy"]]
-    )
+## Inspect model's performance
+
+It was already shown how to obtain the performance of the model. However,
+those results are valid for a particular 80/20 split  of train and test
+data(determined by the seed used), and thus it represents the performance of
```
TuomasBorman: data(determined by the seed used) --> add space

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+                                  seed = seed,
+                                  kfold = 2, 
+                                  cv_times = 2, 
+                                  training_frac = .8)
+                           }
+                   )
+
+# Join the 2 new rf outputs with the first output generated
+rf_list <- multiple_rf
+rf_list[[3]] <- rf
```
TuomasBorman: Does this work
`
rf_list <- c(rf_list, lrf)`

inst/pages/machine_learning.qmd
TuomasBorman: And the line width should be max 80 characters

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+                   shape = 21,
+                   show.legend = FALSE,
+                   alpha = 0.65) +
+        stat_summary(geom = "pointrange",
+                     fun.data = mean_se, 
+                     color = "black", 
+                     show.legend = FALSE) +
+        scale_y_continuous(limits = c(0.5, 1)) + 
+        theme_classic()
+
```
TuomasBorman: Can this be somehow simplified? This is rather complex and hard to follow for ordinary user

inst/pages/machine_learning.qmd
TuomasBorman: I wonder if we should direct user to mikropml documentation instead of implementing this multimodel training to avoid duplication. Or does this include information that is not included in here: https://www.schlosslab.org/mikropml/articles/parallel.html

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
 
-Our model shows a slight improvement over previous versions, but its
-performance remains average. The key to better performance lies in the
-quality of the data; even the most advanced models cannot compensate for
-data that lacks essential information. The dataset do not have sufficient
-amount of data on carcinoma to build an accurate predictive model.
+This plot shows the features with the highest effect on model's performance,
+or in other words, which features have most of the information for
+predicting the outcome. Notice that within the most influential features
+there are some grouped variables.
```
TuomasBorman: We could emphasize more the interpretation of results. What are the most important features? What this means?

TuomasBorman: Also in ROC plot: What the ROC plot shows. What can we say based on the ROC plot?

TuomasBorman (changes requested):
Looks good! Check the comments.

I am wondering if we should more clearly separate the topics to

1. Background
2. Preprocessing
3. Training
    - Binary classification
    - Regression task: for instance age (for instance data(hitchip1006, package ="miaTime"), even though the data is not 16S or shotgun)
4. Model metrics
   - Typical model metrics for classification, also say couple words about multiclass
   - Metrics for regression
5. Visualization
   - Classification
   - Regression
 6. Feature importance

inst/pages/machine_learning.qmd
Benjamin-Valderrama: In added `|# output : FALSE` as part of the code chunk options, so the output of that `table` call won't be printed (at least it didn't when rendering locally). Do you wan't me to remove the comment and allow the outputs to be shown then?

TuomasBorman: Sorry, I did not look carefully enough. I think you could print the table. It might be easier if we decide to change the dataset at some point.

Same thing also with the text. If there are specific interpretations from the results, e.g., "this bacteria x is the most important feature in prediction", the "x" could be inline code. It would automatically update for new dataset.

inst/assets/bibliography.bib
```diff
@@ -2276,6 +2276,61 @@ @article{Rahman2023
...
+	author = {Qin, Junjie and Li, Yingrui and Cai, Zhiming and Li, Shenghui and Zhu, Jianfeng and Zhang, Fan and Liang, Suisha and Zhang, Wenwei and Guan, Yuanlin and Shen, Dongqian and Peng, Yangqing and Zhang, Dongya and Jie, Zhuye and Wu, Wenxian and Qin, Youwen and Xue, Wenbin and Li, Junhua and Han, Lingchuan and Lu, Donghui and Wu, Peixian and Dai, Yali and Sun, Xiaojuan and Li, Zesong and Tang, Aifa and Zhong, Shilong and Li, Xiaoping and Chen, Weineng and Xu, Ran and Wang, Mingbang and Feng, Qiang and Gong, Meihua and Yu, Jing and Zhang, Yanyan and Zhang, Ming and Hansen, Torben and Sanchez, Gaston and Raes, Jeroen and Falony, Gwen and Okuda, Shujiro and Almeida, Mathieu and LeChatelier, Emmanuelle and Renault, Pierre and Pons, Nicolas and Batto, Jean-Michel and Zhang, Zhaoxi and Chen, Hua and Yang, Ruifu and Zheng, Weimou and Li, Songgang and Yang, Huanming and Wang, Jian and Ehrlich, S. Dusko and Nielsen, Rasmus and Pedersen, Oluf and Kristiansen, Karsten and Wang, Jun},
+	journal = {Nature},
+	volume = {490},
+	pages = {55--60},
+	year = {2012},
+	publisher = {Nature Publishing Group US New York},
```
antagomir: double curly brackets {{}} to keep capitalization?

inst/assets/bibliography.bib
```diff
@@ -2276,6 +2276,61 @@ @article{Rahman2023
...
+	doi = {10.1038/nature11450}
+}
+
+@article{mikropmlSoftware,
```
antagomir: we prefer using "lastnameyear" like "topguoclu2021" as the bibkey, would be good to update this here and elsewhere. It will also help to spot duplicate refs.

inst/assets/bibliography.bib
```diff
@@ -2276,6 +2276,61 @@ @article{Rahman2023
...
+	title = {{mikropml}: User-Friendly R Package for Supervised Machine Learning Pipelines},
+	author = {Begüm D. Topçuoğlu and Zena Lapp and Kelly L. Sovacool and Evan Snitkin and Jenna Wiens and Patrick D. Schloss},
+	journal = {Journal of Open Source Software},
+	year = {2021},
+	volume = {6},
+	number = {61},
+	pages = {3073},
+	doi = {10.21105/joss.03073},
+	url = {https://joss.theoj.org/papers/10.21105/joss.03073}
```
antagomir: doi is enough?

inst/assets/bibliography.bib
```diff
@@ -2276,6 +2276,61 @@ @article{Rahman2023
...
+} 
+
+@article{mikropmlApplied,
```
antagomir: lastnameyear for bibkey?

inst/assets/bibliography.bib
```diff
@@ -2276,6 +2276,61 @@ @article{Rahman2023
...
+	title = {A Framework for Effective Application of Machine Learning to Microbiome-Based Classification Problems},
+	author = {Begüm D. Topçuoğlu and Nicholas A. Lesniak and Mack T. Ruffin and Jenna Wiens and Patrick D. Schloss},
+	journal = {mBio},
+	year = {2020},
+	volume = {11},
+	number = {3},
+	pages = {10.1128/mbio.00434-20},
+	doi = {10.1128/mbio.00434-20},
+	url = {https://journals.asm.org/doi/abs/10.1128/mbio.00434-20}
```
antagomir: doi is enough? (check also for other refs)

inst/assets/bibliography.bib
```diff
@@ -2276,6 +2276,61 @@ @article{Rahman2023
...
+	year = {2024},
+	volume = {21},
+	number = {8},
+	pages = {1444--1453},
+	doi = {10.1038/s41592-024-02362-y},
+	url = {https://doi.org/10.1038/s41592-024-02362-y}
+}
+
+@book{james2013_ISLR,
+	title = {An introduction to statistical learning : with applications in {R}},
```
antagomir: remove space before " : "

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+
+# Prepare data for plotting
+top_imp_features_df <- rf_feat_imp |> 
+        slice_max(order_by = perf_metric_diff, 
+                  n = 10) |> 
+        mutate(feat = fct_reorder2(feat, 
+                                   feat, 
+                                   perf_metric_diff, 
+                                   .desc = FALSE))
+
```
antagomir: Any chance to simplify (for casual readers)?

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,509 @@
...
+# Plot feature's effect in models performance
+top_imp_features_df |> 
+        ggplot(aes(x = perf_metric_diff, y = feat)) +
+        geom_col() +
+        scale_x_continuous(
+                limits = c(0, 
+                           top_imp_features_df$perf_metric_diff[1] * 1.1),
+                expand = c(0, 0)) +
+        labs(x = "Change in performance",
```
antagomir: this, too

antagomir (commented):
Looks good! I agree with comments from Tuomas & added some other small ones.

Benjamin-Valderrama:
Thank you for all your comments, I appreciate your insights regarding the content and structure of the chapter.

I've just pushed the changes to address the issues you flagged. I've added the content you suggested, and removed sections to avoid redundancy with other resources. I've also extended the interpretation of the results (both technically and biologically), and also changed the overall structure of the chapter to better reflect what @TuomasBorman asked for in the first comment made. 

Regarding the use of the hitchip1006 dataset in regression tasks suggested by @TuomasBorman, I decided to use BMI of each participants instead. I made that decision to avoid deviations to topics other than the gut microbiome. Notice that I added biological justification as per way it could make sense to explore associations between gut microbiome and BMI.

I hope this changes make sense and addressed your concerns. 

Thanks,
Ben

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+To do so, the R package `mikropml`[@topcuoglu2020;@topcuoglu2021] will be
+used throught this chapter. This package was developed to offer a 
+user-friendly interface to supervised ML algorithms implemented in the
+`caret` package. 
+[Here](https://www.schlosslab.org/mikropml/articles/introduction.html#the-methods-we-support)
+is a list of models well supported by mikropml and
+[here](https://topepo.github.io/caret/available-models.html)
+a complete list of other models available in caret that *might* work when
+called through mikropml.
+
```
TuomasBorman: I think this "might work" thing could be removed. I think it is enough that we provide the link to mikropml and its models

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+    find_feature_importance = FALSE
 )
+```
+
+For regression tasks, each tree predicts a numerical outcome (e.g.,
+BMI) based on the values of the features (e.g., microbial taxa). Each 
+tree split the data to minimize the difference between predicted and 
+real (observed) values, often using metrics like mean squared 
+error (MSE). One strength of RF is that it can learn non-linear patterns 
+in the data that other regression algorithms might not capture.
```
TuomasBorman: This feels little bit that learning non-linear relationships is special for RF.

You could rephrase this little bit. How about something like:

"... that other simpler regression algorithms ...

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
 
-# Get predictions
-res <- model[["pred"]]
-# Print result
-confusionMatrix(data = res[["pred"]], reference = res[["obs"]])
+Regarding its implementation, notice that just by changing the outcome 
+from a categorical to a continuous variable, RF models can now be used
+in regression tasks:
+```{r}
```
TuomasBorman, line 341: Add single new line above code chunk

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+    cv_times = 2, 
+    training_frac = .8,
+    find_feature_importance = FALSE
+)
+```
+
+Although not demonstrated in the chapter, XGBoost can be used
+in regressions tasks using similar code to the one shown above.
+The only difference will be the outcome variable selected.
+See the RF example for reference.
```
TuomasBorman, line 393: This is already said in lines 361-362. I suggest modifying the lines 361-362 so that they do not overlap with this paragraph

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+
+
+## Model performance metrics
+
+Under the hood, the function `run_ml` generates a 80/20 split of 
+the data to train and test a model of interest. That means
```
TuomasBorman: This can be probably adjusted. Add: "by default" or something

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+iterations. This gives a more accurate assessment of the model's 
+performance. 
+
+That approach is discussed in the 
+[documentation](https://www.schlosslab.org/mikropml/articles/parallel.html) 
+of `mikropml`. Therefore, we will focus in interpreting the results
+of one split. The goals below are to discuss the different performance 
+metrics, to compare metrics from two models used in the classification 
+of the same observations, and to contrasts the results of these models
+with previous analyses in the published literature.
```
TuomasBorman: I think these paragraphs can be improved by combining them differently.

You could just say what is the goal below. The you could move this multi-split technique completely to "tip callout block". I think it would make the text more clear.

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+Metrics of the RF model:
+``` {r}
+#| label: 5.1 - RF classification performance
+
+# RF performance in classification tasks
+rf_classification[["performance"]] |> knitr::kable()
+```
+
+Metrics of the XGBoost model:
+``` {r}
```
TuomasBorman, line 444: Add new line before the chunk. Also in line 430

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+
+# XGBoost performance in classification tasks
+xgb[["performance"]] |> knitr::kable()
+```
+
+A common metric to quickly assess model performance is the area under 
+the receiver operator characteristic curve 'AUC'. Notice that
+the result tables include two types of AUC metrics: `cv_metric_AUC`,
+which represents AUC for the 80% of data used in training, and `AUC`, 
+which is the AUC for the 20% used in testing.
```
TuomasBorman: How about multi-class? The standard AUC is for binary. There are some extensions to AUC for multi-class but are they included in mikropml?

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+the receiver operator characteristic curve 'AUC'. Notice that
+the result tables include two types of AUC metrics: `cv_metric_AUC`,
+which represents AUC for the 80% of data used in training, and `AUC`, 
+which is the AUC for the 20% used in testing. 
+
+Researchers are often interested in scenarios where `AUC` is as good as 
+or better than `cv_metric_AUC`. The opposite suggests that the model may 
+be overfitted, meaning that the model fails to generalize to new data.
+Notice, however, that small drops of model's performance in the test 
+compared to the train data are acceptable and shouldn't be a concern.
```
TuomasBorman: If the AUC for unseen data is much grater than AUC for CV, doesn't this indicate that for some reason the test set includes easier samples to predict? This might be due to failed splitting to train and test set.

Usually, it is expected that AUC-CV is greater than AUC for test data as AUC-CV includes samples for which the model was optimized. If it is much greater, then it incicates over-fitting.

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+Now let's take a look at the results. First, notice that AUC values 
+suggest a relatively good performance considering that predicting 
+outcomes based on complex microbiome data is typically a challenging 
+task. In addition, notice that the performance of RF and XGBoost are
+similar (RF AUC = 
+`{r} round(rf_classification[["performance"]][["AUC"]], 4)` and
+XGBoost AUC = `{r} round(xgb[["performance"]][["AUC"]], 4)`). This 
+illustrates an important point in ML: sometimes simpler models can
+perform as good as more complex ones. Thus, it is often a good idea 
+to train different models and compare them to each other.
```
TuomasBorman: This also illustrates that the information is in the data. Even the best possible model cannot learn information that is not in the data.

inst/pages/machine_learning.qmd
TuomasBorman: Noticed that you are discussing this below. Here you might want to add that these examples are for binary classification

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+#| label: 5.2 - XGBoost classification performance
+
+# XGBoost performance in classification tasks
+xgb[["performance"]] |> knitr::kable()
+```
+
+A common metric to quickly assess model performance is the area under 
+the receiver operator characteristic curve 'AUC'. Notice that
+the result tables include two types of AUC metrics: `cv_metric_AUC`,
+which represents AUC for the 80% of data used in training, and `AUC`,
```
TuomasBorman: You could add somewhere in model training what cross-validation means. It seems that CV is not covered. Moreover, sometimes we have so little data that we cannot have separate training and tests sets so we use only cross-validation

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+    geom_hline(color = "grey30", linetype = "dashed", yintercept = 0.5)
+# 2.2 Add axis titles and custom theme
+prc_p <- prc_p +
+    labs(x = "Recall or Sensitivity", y = "Precision", title = "PRC") +
+    theme_bw()
+
+
+# 3. Plot ROC and PRC side by side
+roc_p + prc_p + plot_layout(guides = "collect")
+```
```
TuomasBorman: Was there plotting functions in mikropml package?

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+and regression tasks. However, researchers are often interested in
+generating visual representations of the performance of their models.
+Thus, the goal of this section is to show code to create those
+visualizations, as well as providing insights in their interpretations. 
+
+### Classification
+
+The area under the receiver-operator characteristic (ROC) curve,
+also called 'AUC' was introduced before as common metric to assess 
+model's performance in classification tasks. As it name suggests,
```
TuomasBorman: You could again say that this is for binary classification. Also add somewhere link to multiclass visualization

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+    geom_point(aes(x = pred, y = BMI, color = diagnosis), size = 2)
+# Add line showing perfect prediction
+obs_vs_pred <- obs_vs_pred + geom_abline(color = "grey30", linetype = "dashed")
+# Make axes of same scale
+obs_vs_pred<- obs_vs_pred + coord_equal(xlim = c(17,30), ylim = c(17,30))
+# Custom axes titles
+obs_vs_pred <- obs_vs_pred + labs(x = "Predicted BMI", y = "Observed BMI")
+# Add custom theme and visualize the plot
+obs_vs_pred + theme_bw()
 ```
```
TuomasBorman: Same thing: wast there visualization function in mikropml?

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+(2) T2D patients with obesity, and (3) T2D patients without obesity by 
+integrating BMI information. In such cases, metrics like AUC (developed 
+for binary classification) can be generalized, like in the case of 
+multiclass AUC, also referred to as One-vs-Rest AUC. This approach 
+evaluates how well each class is distinguished from the others, 
+allowing the assessment of model performance in multiclass problems.
+Multiclass classification can be implemented with `run_ml` too, as 
+described 
+[here](https://www.schlosslab.org/mikropml/articles/introduction.html?q=multiclass#multiclass-data).
+
```
TuomasBorman, line 500: You could consider adding a confusion matrix. That works also for multiclass

inst/pages/machine_learning.qmd
TuomasBorman: This chunk s simple for one that understand what these does, but for reader this kind of code-heavy chunk might be hard to grasp

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+This plot clearly indicates a poor performance of our model. This
+usually means that either the gut microbiome is not related to BMI, 
+making it impossible for the model to use microbiome patterns to predict
+BMI, or that our model is overfitted. Since the RMSE for training 
+(`{r} rf_regression[["performance"]][["cv_metric_RMSE"]]`) and test 
+(`{r} rf_regression[["performance"]][["RMSE"]]`) data are similar, 
+we may conclude that gut microbiome composition is not very 
+informative of BMI.
+
+
```
TuomasBorman, line 770: Remove extra line

inst/pages/machine_learning.qmd
```diff
@@ -4,253 +4,817 @@
...
+Note that both groups show a similar spread in BMI.
+
+This plot clearly indicates a poor performance of our model. This
+usually means that either the gut microbiome is not related to BMI, 
+making it impossible for the model to use microbiome patterns to predict
+BMI, or that our model is overfitted. Since the RMSE for training 
+(`{r} rf_regression[["performance"]][["cv_metric_RMSE"]]`) and test 
+(`{r} rf_regression[["performance"]][["RMSE"]]`) data are similar, 
+we may conclude that gut microbiome composition is not very 
+informative of BMI.
```
TuomasBorman: It seems that the model is practically useless for BMI, so I suggest removing "very" (I think this is still good example to show that sometimes the model cannot predict the outcome)

TuomasBorman (changes requested):
I like this new style

inst/pages/machine_learning.qmd
Benjamin-Valderrama: I think that is implicitly acknowledged when describing the dataset and diagnoses (see the Setup section). However, I have mentioned earlier that AUC is used for binary classifications by adding some words to the lines you flagged, instead of waiting to almost the end of the section.

inst/pages/machine_learning.qmd
Benjamin-Valderrama: > 'Moreover, sometimes we have so little data that we cannot have separate training and tests sets so we use only cross-validation'

That was mentioned in the current text (see 24.1.2 Supervised machine learning). 

Maybe I can add a small explanation on what CV is in that paragraph.

inst/pages/machine_learning.qmd
Benjamin-Valderrama: I agree with you and I also think this is relevant to mention. Indeed, I did it in the results of the regression task, where the performance is very bad and I can add your phrasing in there too in the new version. However, I think in this situation that is not what happens here. Some patterns are there and they can be learned by the models (indeed, it seems like they did based on their performances). However, RF, which is simpler than XGBoost, achieved a similar performance in this mock example. In the regression task, however, it seems like gut microbiome holds no relevant information to predict BMI, which would be the case you alluded to in your comment.

inst/pages/machine_learning.qmd
Benjamin-Valderrama: Thanks, I can see the problem with that description. I've changed that paragraph for regression and classification.

inst/pages/machine_learning.qmd
Benjamin-Valderrama: Not sure if you meant a link to a tutorial on multiclass visualization or a 'real-world' application of that using gut microbiome data. I added the later. Please tell me if you had other option in mind.

inst/pages/machine_learning.qmd
Benjamin-Valderrama: You are right, there are plotting functions. I used them in the previous PR I made where I was running the same type of ML model with multiple seeds and calculating the mean performance across iterations. However, after addressing your suggestion of focussing in just one iteration, I faced the following situation when using the plotting functions from `mikropml`.

By using their plotting functions, you can see that confidence intervals are displayed in the the ROC curves despite that sensitivity and specificity were calculated for just one model:

```
rf_senspec |> # Generated in section '24.6.1 Classification'
        mikropml::calc_mean_roc() |> 
        mikropml::plot_mean_roc()
```
image

This is a consequence of the behaviour of `calc_mean_roc()`. To calculate the mean sensitivity it groups values by specificity. Since the function `calc_model_sensspec()` (used to generate the object `rf_senspec`) returns a dataframe that may have multiple sensitivities for the same specificity value, the plotted mean roc has those broken intervals shown in the figure above. None of those functions receives arguments to modify their behaviour in this regard, so I think there isn't much to do with this workflow. 

A small fix to that would be to force the upper and lower bounds to be equal to the mean sensitivity across all specificity values, which will generate a plot without CI:

```
rf_senspec |> 
        calc_mean_roc() |> 
        mutate(upper = mean_sensitivity,
               lower = mean_sensitivity) |>
        plot_mean_roc()
```
image
However, I think that solution may induce confusions in the readers, as it adds lines just to fix an artifact likely generated by using a function for a purpose different to what it was developed for in the first time. My reason to believe this is that examples where multiple seeds are used for the same type of model are prevalent in the docs of `mikropml`.  Additionally, I don't think that explaining the 'fix' or the context of multiple-vs-single seed models in the main text of the chapter can make things clearer.

It is true that my solution in the current PR is lengthy but it is commented step by step, which I think helps readers to understand the process of building the visualizations. This was the rational behind my decision but I'd appreciate your thoughts.

inst/pages/machine_learning.qmd
Benjamin-Valderrama: See my answer to your similar comment above. Additionally, as far as I understand from the documentation, `mikropml` doesn't provide helper functions for the visualization of the performance of a single regression model. Definitely not like this observed-vs-predicted plot at least. 

Finally, I truly believe the comments are clear in terms of indicating the reasons of each line of code.  However, if worried about this chunk being hard to grasp, maybe it can be broken and new text added to better clarify what is happenning (and why).

inst/pages/machine_learning.qmd
Benjamin-Valderrama: Sure. It was in an ironic sense. It was removed and I added your previous suggestion:

> This also illustrates that the information is in the data. Even the best possible model cannot learn information that is not in the data.

below that paragraph.

inst/pages/machine_learning.qmd
TuomasBorman: I think we could suggest mikropml developers to crate a function for creating ROC for single model. If you agree, can you suggest?

Until then,your "custom script" is good. It is well-commented and clear example

inst/pages/machine_learning.qmd
TuomasBorman: Sorry for being unclear. By point was that now we only show how to visualize binary models. If reader has multiclass problem, then we should provide further reading for visualizing multiclass models. I think tutorial would be the best, preferably, we could provide reference for to R package that does plots easily.

inst/pages/machine_learning.qmd
Benjamin-Valderrama: I think they are not very active in updating the code or adding new features. I saw you opened an issue and then a PR some weeks ago. I also opened an issue some days after you but none got attention. Also, the last comit was almost 2 years ago.

TuomasBorman: It seems so, but the main developer has been prepared couple PRs just few weeks ago

TuomasBorman (approved):
Looks good!! Does @antagomir have further comments?

antagomir:
I think we can merge now, this seems very useful. More revisions can be discussed later shall needs arise.

PR #678, opened 2025-03-10
Docker

artur-sannikov:
Let me know if the explanation is clear.

To successfully build the book, we also need to install miniconda, python and mofapy2 during the book build process.


inst/pages/multiassay_ordination.qmd
```diff
@@ -54,6 +54,9 @@ remove them.
 #| label: mofa3
 
 library(MOFA2)
+reticulate::install_miniconda(force = TRUE, update = FALSE)
+reticulate::use_miniconda(condaenv = reticulate::miniconda_path(), required = TRUE)
+reticulate::py_install(packages = c("mofapy2"), pip = TRUE, python_version=3.12)
```
TuomasBorman: Add these to another chunk and explain why these are needed. Moreover, I think the installation should be disabled if they are already installed (so no update).

We expect that often users just copy-paste this chunk  without thinking. This update can cause problems, especially in failure, since users expect that this should be run every time.

README.md
```diff
@@ -40,6 +40,47 @@ devtools::install('.')
...
+docker run --volume ./:/project -p 8787:8787 -e PASSWORD=1234 ghcr.io/microbiome/oma
+```
+
+The command above will start the image based on the Biocoductor docker image, which includes RStudio Server.
+The `--volume` maps the path on your host (the machine you are using) to the path inside the container.
+In the command, it maps the current working directory to the `/project` directory inside the container.
+If the current directory is the book's repository, it will be available inside the container under `/project`.
+Thus, in this case, the version of the book coincides with the bleeding-edge version available in the repo.
+If you do not specify this option, the book Quarto files are still available in the `/opt/pkg` directory
+inside the container, but any changes you make to the files, will be _discarded_ when you stop the container.
```
TuomasBorman: Add line breaks when the line is over 80 characters long (it seems that some might be).

TuomasBorman (changes requested):
Thanks, looks good! Check comments

README.md
```diff
@@ -40,6 +40,47 @@ devtools::install('.')
 BiocBook::preview(BiocBook::BiocBook('.'))
 ```
 
+### Docker installation
+
+Pre-built [Docker images](https://github.com/microbiome/OMA/pkgs/container/oma) are avaialble for an easy deployment.
+
+1. Install Docker for your platform: [Linux](https://docs.docker.com/engine/install/),
+[Mac](https://docs.docker.com/desktop/setup/install/mac-install/), [Windows](https://docs.docker.com/desktop/setup/install/windows-install/).
+2. Pull the image from our repository:
```
TuomasBorman: You could specify terminal

inst/pages/multiassay_ordination.qmd
artur-sannikov: Just pushed, does it look better?

README.md
artur-sannikov: What do you mean?

TuomasBorman: Sorry. As many of the readers have microbiology background "sh" might not be clear for everyone.  By specifying "Pull the image from our repository by running the following command in terminal" is more explicit

I will check the instructions in more detail during evening (it seems to take some time to pull the docker)

README.md
```diff
@@ -40,6 +40,56 @@ devtools::install('.')
...
+
+```sh
+docker pull ghcr.io/microbiome/oma:latest
+```
+
+You can also specify any of the available tags to pull a particular version of
+the image.
+
+3. Switch to the directory of the book, for example, with `cd` command.
+
```
TuomasBorman: Thanks! I was able to load the image and open it. This part is not clear, and it should be clarified.

You can link to instruction on how to clone the repo. Alternatively, as many of the readers might not be familiar with git, you might instruct readers to load the repo from GH and extract the files.

inst/pages/multiassay_ordination.qmd
```diff
@@ -50,6 +50,22 @@ duplicates are present. We can add
...
 remove them.
 
+To run MOFA2, it needs to connect `mofapy2` Python packages via the `reticulate`
+package, as clarified in the [MOFA2 vignette](https://bioconductor.org/packages/release/bioc/vignettes/MOFA2/inst/doc/getting_started_R.html#6_Build_and_train_the_MOFA_object).
+
+The following chunk does not install and does not update `miniconda` if already
+installed. Furthermore, it does not install the latest version of `mofapy2`.
+If this setup causes issues, update `miniconda` and `mofapy2`.
+
+```{r, message=FALSE, warning=FALSE}
```
TuomasBorman: Use quarto formatting as done with label

inst/pages/multiassay_ordination.qmd
```diff
@@ -50,6 +50,22 @@ duplicates are present. We can add
...
+#| label: mofapy2
+reticulate::install_miniconda(force = FALSE, update = FALSE)
+reticulate::use_miniconda(condaenv = reticulate::miniconda_path(), required = FALSE)
+reticulate::py_install(packages = c("mofapy2"), pip = TRUE,
+                       pip_ignore_installed = TRUE, python_version=3.12)
+```
```
TuomasBorman: Instead format like this

```
reticulate::py_install(
    packages = c("mofapy2"), pip = TRUE,
    pip_ignore_installed = TRUE, python_version=3.12
)
```

If the line is too long

inst/pages/multiassay_ordination.qmd
```diff
@@ -50,6 +50,22 @@ duplicates are present. We can add
...
+
+
```
TuomasBorman: Remvoe extra line

inst/pages/multiassay_ordination.qmd
```diff
@@ -50,6 +50,22 @@ duplicates are present. We can add
 `rownames(mae[[1]]) <- getTaxonomyLabels(mae[[1]], make.unique=TRUE)` to
 remove them.
 
+To run MOFA2, it needs to connect `mofapy2` Python packages via the `reticulate`
+package, as clarified in the [MOFA2 vignette](https://bioconductor.org/packages/release/bioc/vignettes/MOFA2/inst/doc/getting_started_R.html#6_Build_and_train_the_MOFA_object).
+
+The following chunk does not install and does not update `miniconda` if already
+installed. Furthermore, it does not install the latest version of `mofapy2`.
+If this setup causes issues, update `miniconda` and `mofapy2`.
```
TuomasBorman: This looks good

README.md
artur-sannikov: I wrote the instructions regarding `git clone`, but don't know how to handle the `cd` part. Although Docker is easy to use, it assumes some familiarity with basics of Unix systems, like `cd` command, so should we delve deep here or suggest the readers to familiarize with the command by themselves? I'd like to achieve a good balance between bigger picture and the details.

inst/pages/multiassay_ordination.qmd
artur-sannikov: I don't follow. I copied the style of the other cells. I only added the space between the label and the code now.

TuomasBorman: There are chunks that do not yet follow the quarto format. It is better to follow the new format when chunks are modified.

Something like this

```
```{r}
#| message = false
#| warning = false

```

inst/pages/multiassay_ordination.qmd
```diff
@@ -50,8 +50,30 @@ duplicates are present. We can add
...
+
+reticulate::install_miniconda(force = FALSE, update = FALSE)
+reticulate::use_miniconda(condaenv = reticulate::miniconda_path(), required = FALSE)
+reticulate::py_install(
+    packages = c("mofapy2"), pip = TRUE,
+    pip_ignore_installed = TRUE, python_version=3.12
+)
+```
+
+```{r}
```
TuomasBorman: Can the installation be done just by specifying  `use_basilisk = TRUE`

https://rdrr.io/bioc/MOFA2/man/run_mofa.html

TuomasBorman, line 73: Can the installation be done just by specifying  `use_basilisk = TRUE`

https://rdrr.io/bioc/MOFA2/man/run_mofa.html

README.md
TuomasBorman: I think it is good now

inst/pages/multiassay_ordination.qmd
artur-sannikov: Theoretically, yes, but it does not seem to work. See [this issue](https://github.com/bioFAM/MOFA2/issues/48).

TuomasBorman: Alright, thanks!

I noticed that running this chunk again leads to an error:

```
Error: Miniconda is already installed at path "~/.local/share/r-miniconda".
- Use `reticulate::install_miniconda(force = TRUE)` to overwrite the previous installation.
```

TuomasBorman (changes requested):

inst/pages/multiassay_ordination.qmd
TuomasBorman: The most optimal solution would be to check whether these are installed and run them only if they are not.

artur-sannikov: The way to do that does not exist in the `reticulate` package. The only way I can thing of is to check whether the path `~/.local/share/r-miniconda` exists, but we cannot guarantee that it's installed there.

So, the `force = TRUE` option is the best, IMHO. What's your opinion?

TuomasBorman: How about: reticulate:::miniconda_exists()

https://rdrr.io/github/rstudio/reticulate/src/R/miniconda.R

artur-sannikov: It's a good idea!

However, when rerunning the same code I get

```
! AttributeError: module 'mofapy2' has no attribute 'run.entry_point'
Run `reticulate::py_last_error()` for details.
Backtrace:
    ▆
 1. └─MOFA2::run_mofa(model, use_basilisk = TRUE)
 2.   └─basilisk::basiliskRun(...)
 3.     └─MOFA2 (local) fun(...)
 4.       └─MOFA2:::.run_mofa_reticulate(object, outfile, save_data)
 5.         ├─mofa$run.entry_point
 6.         └─reticulate:::`$.python.builtin.module`(mofa, "run.entry_point")
 7.           └─reticulate::py_get_attr(x, name, FALSE)
                                                                                                       



Quitting from lines 142-152 [mofa6] (multiassay_ordination.qmd)
Execution halted
```

artur-sannikov: So, we need to force the reinstall...

TuomasBorman: Should the `use_basilisk` be `FALSE`

TuomasBorman: Or does that come from line 67?

TuomasBorman: The force installation led to an error for me. I am not sure why, but it seems that we cannot use force installation.

artur-sannikov: What error?

TuomasBorman: > Alright, thanks!
> 
> I noticed that running this chunk again leads to an error:
> 
> ```
> Error: Miniconda is already installed at path "~/.local/share/r-miniconda".
> - Use `reticulate::install_miniconda(force = TRUE)` to overwrite the previous installation.
> ```

artur-sannikov: Yep, exactly. And if I run it without it I get this [issue](https://github.com/bioFAM/MOFA2/issues/48).

TuomasBorman: Does it now work?

artur-sannikov: If I rerun the book build, I get into this error without `force = TRUE`.

TuomasBorman: So if you now add force = TRUE it works?

The checks should prevent updating these packages when user runs the chunk

artur-sannikov: Okay, I tested it  several more times completely from scratch. Now everything works as expected no matter how many times I rerun it. Could you try it now?

artur-sannikov: Thanks for your contribution. I think these lines fixed it:

```R
if( !condaenv_exists(miniconda_path()) ){
    use_miniconda(condaenv = miniconda_path(), required = FALSE)
}

# Check if mofapy2 is installed, and install it if not
if( !py_module_available("mofapy2") ){
    py_install(
        packages = "mofapy2",
        pip = TRUE,
        pip_ignore_installed = TRUE,
    )
}
```

TuomasBorman: Seems to work

TuomasBorman (approved):
Thanks!

PR #682, opened 2025-03-11
Add caution for Jaccard index

### LINKED ISSUE #681: Jaccard issue
It was reported by @nuorenarra that the _vegan_ "bray" dissimilarity default does not correspond to the standard version of the widely known Bray-Curtis dissimilarity. 

TODO: let us consider how the defaults should be set in mia (overriding vegan defaults?). At least this is something to mention in the documentation and OMA.

TuomasBorman:


antagomir:
Just wondering if we should make this binary (classical) version the default..? To me that would make sense.

nuorenarra:
Yes I agree. And perhaps we can mention this in the OMA book somewhere. I can add a sentence somewhere relevant as I go through the book.

TuomasBorman:
Is there "real" use-case for Jaccard index with `binary = FALSE`? It might also be problematic if we choose other default value than vegan does. Although, I guess it is ok in this case

If Jaccard `binary = TRUE` should be used always, then it might be easier choice

antagomir:
Ok I added an issue to mia on this. Let's be careful not to forget ongoing discussions with closed PRs.

nuorenarra:
in vegan (and subsequently in phyloseq) the default Jaccard is with binary= FALSE, meaning that it is an abundance weighted metric. vegdist documentation says "Jaccard index is computed as 2B/(1+B), where B is Bray–Curtis dissimilarity. ". Note that if you read the Wikipedia page of Jaccard Index /Jaccard distance, it explicitly says it is a metric of "similarity of binary attributes". Also in ecology text books Jaccard is a binary metric, so an average biologist will hardly notice that the default vegdist Jaccard is actually an asundance-weighted metric. This is why I suggest we would use Jaccard, binary=T as the default Jaccard or make a very clear note of this discrepancy where we introduce dissimilarity (beta diversity) metrics

antagomir:
See https://github.com/microbiome/mia/issues/706

PR #692, opened 2025-03-20
Improve preprocessing chapters

### LINKED ISSUE #690: Improve filtering and transformation chapters
I got great questions from one person that participated our course earlier. I noticed that we have not emphasized this enough in OMA. These are so basic things so we have to improve related chapters.

**1. Is filtering done prior to transformation?**

The person wanted to do prevalence filtering and apply CLR transform. He wandered what is the correct order; should filtering be applied first or after transformation.

Generally, filtering is done first and then transformation is applied (this is also why the book chapters are in this order). However, there are some filtering methods that require normalized data.

This paper has nice flow diagram, we could cite that:

https://pmc.ncbi.nlm.nih.gov/articles/PMC10569954/#fig0010

**2. Should CLR be applied to coutns or relative abundances?**

As CLR is scale-invariant, in theory it does not matter whether to apply the transformation to relative abundances or counts. However, in practice it depends because pseudocount introduces bias.

```
library(mia)
data("GlobalPatterns")
#
tse <- GlobalPatterns
tse <- transformAssay(tse, method = "relabundance")
tse <- transformAssay(tse, method = "relabundance", name = "relabundance2", pseudocount = TRUE)
#
tse <- transformAssay(tse, method = "clr", pseudocount = TRUE, name = "clr1")
tse <- transformAssay(tse, method = "clr", assay.type = "relabundance", pseudocount = TRUE, name = "clr2")
tse <- transformAssay(tse, method = "clr", assay.type = "relabundance2",  name = "clr3")
#
mat <- assay(tse, "clr1") - assay(tse, "clr3")
mat2 <- assay(tse, "clr2") - assay(tse, "clr3")
#
max(mat)
[1] 4.440892e-15
max(mat2)
[1] 3.149329
```

In example above:

- CLR1: pseudocount -> CLR
- CLR2: relative transformation -> pseudocount -> CLR
- CLR3: pseudocount -> transformation -> CLR

CLR1 == CLR3, however, if pseudocount is applied to relative abundances, the result differs.

We should make decision on what we recommend in OMA: should CLR be applied to counts or relative abundances? Few points that came to my mind:

a) CLR should be applied to "raw" values

- CLR transformation does not include relative transformation in its definition.
- In one sense, it is more clear to add CLR directly to raw values without additional steps.

b) CLR should be applied always to relative abundances

- Some pipelines output only relative abundances (Metaphlan). Relative transformation can always be applied, meaning that we could standardize this approach (always apply relative transformation prior to CLR)
- Another question is, should pseudocount be applied to relative transformation (CLR3 in example above) or CLR transformation step (CLR2)? CLR3 approach could be standardized despite on the type of raw values. However, I think pseudocount "should" be applied as early as possible (to raw values).

No strong opinion, but I think a) is more clear. Anyways, this should be clarified in the book.

@antagomir 

-Tuomas

TuomasBorman:
Related to this issue: https://github.com/microbiome/OMA/issues/690


inst/assets/_book.yml
```diff
@@ -34,15 +34,16 @@ book:
...
+        - pages/preprocess.qmd
+        - pages/quality_control.qmd
         - pages/subsetting.qmd
         - pages/agglomeration.qmd
         - pages/transformation.qmd
-    - part: "Exploration & QC"
-      chapters:
-        - pages/quality_control.qmd
-        - pages/composition.qmd
     - part: "Diversity & similarity"
```
TuomasBorman, line 44: Note that I have changed the organization of these chapters.

Now there are "QC & preprocessing" topic that holds QC, filtering and transformation. Moreover, I added short introduction to this topic (preprocess.qmd)

TuomasBorman: The idea is to briefly describe the context of the topic. If this is good idea we can use this same idea in other topics also. Or we can remove this preprocess.qmd file and add the information to relevant chapters instead.

PR #693, opened 2025-03-21
Linux troubleshooting

artur-sannikov:
I added two sections for package installation troubleshooting for Ubuntu and Fedora. These two tests were done on clean Ubuntu and Fedora VMs, but hopefully they should solve issues also on already used systems.

I also think we need to add a link to this file in README.


TuomasBorman (approved):

TuomasBorman:
Can you add link to readme?

TuomasBorman (approved):

PR #696, opened 2025-03-25
Aura's edits on chapter 9 and 10

nuorenarra:
edits to enhance clarity for biologist readers. 

General notes:

Chapter 9 Agglomeration
 
1.       On row 44: “track the influence of changing conditions through these levels.” 
--> I don’t understand what this means. What are the changing conditions? Are you talking about some variables that may influence the feature distributions? If that is the case, I suggest modify this part to say something like:

 “One of the main applications of taxonomic information in sequencing data
is to agglomerate counts (such as ASV counts) to specific taxonomic levels
and to model how different sample-specific variables may influence the feature composition at different taxonomic levels. For this `mia` contains the `agglomerateByRank()` function.”
 
 
2.      Row 181: “In some cases, both functions may yield the same
results; however, agglomerateByRank() ensures that the entire taxonomy of a
feature is unique for successful agglomeration.

For example, a dataset might contain taxa with the same lower-level rank, even
if their higher-level ranks differ. Thus, `agglomerateByVariable()` should not
be used for aggregating taxonomy ranks.

 Instead, `agglomerateByVariable()` is designed to aggregate data based on other
criteria, such as sample groups or clusters. For instance, we can aggregate the
data by sample types. The function operates similarly to feature aggregation,
but the `by` parameter must be set to "rows."”
  
-->  I don’t understand in which cases these two would yield the same results. If I’ve understood correctly, agglomerateByRank()  considers by default features (rowData), while agglomerateByVariable() considers samples (colData). So how could they yield the same result? I would perhaps delete the part marked red above, as it’s a bit redundant.
 
Chapter 10

1.      In general: This chapter is super important and I think could do with a tiny bit more detail and examples on what each of the provided transformation types mean biologically. I have tried to do this a little bit but more could be added.
 
2.      Row 77: “This simple transformation is relatively widely used in ecological research and has shown good performance in microbiome-based classification performance [@Giliberti2022,
   Karwowska2024].”
 
--> I do not understand the end of this sentence. What is microbiome-based classification performance? I delete this sentence and replaced with an example of when presence absence data is needed. But if you think it’s important, I suggest add it in with an explanation of what it means.
 
The new bit I added:
“This simple transformation is relatively widely used in
   ecological research, whenever we have reasons to be interested in which taxa are
   present in which samples more so than their abundances. An example of such case
   is the process of inferring microbial ecological association networks from covariation
   of taxa across samples. Here, microbes that are rarely seen in the same samples are
   expected to have a negative ecological association (either competitive exclusion with
   each other of different ecological niche preferences). Conversely, looking for negative
   correlations in their relative abundances would not be a meaningful measure of
   ecological associations.”
 
I did not link this to the ecological networks chapter, because I worry they actually do use relative-abundance based covariation as a method of inferring microbial ecological networks there (This is what I criticized earlier). Not sure how to go about this. Perhaps if the network chapters will be revised anyway (and a more commonly used co-occurrence based network method is used as the key example), then this example here is a good one and we can link it to that chapter.

 However, if we wish to keep the network chapters as they are, then this above example is maybe a bit confusing as it says the opposite of what they do there. In this case I suggest replacing it with this other (less good but ok) example:
 
“This simple transformation is relatively widely used in ecological research, whenever we have reasons to be interested in which taxa are present in which samples more so than their abundances. An example of such case is modeling microbiome transmission. Microbiome transmission is assumed to primarily influence which taxa end up where, while their abundance in the system may be more related to within-host processes that influence microbial growth (e.g., host diet, immune system, microbial interactions, etc.). Consequently, it has been shown, that transmission signals are more detectable in presence-absence based measures of microbiome composition compared to when considering abundance-weighted compositions (Raulo et al., 2024).”
 
Raulo, A., Bürkner, P. C., Finerty, G. E., Dale, J., Hanski, E., English, H. M., ... & Knowles, S. C. (2024). Social and environmental transmission spread different sets of gut microbes in wild mice. Nature Ecology & Evolution, 8(5), 972-985.
 
3.       “ 'hellinger' Hellinger transformation is equal to the square root of relative abundances. This ecological transformation can be useful  if we are interested in changes in relative abundances.”
 
--> Could you elaborate a bit what it means that this is useful if we are interested in changes of relative abundance?
 
(Note all these comments and comments for other chapters can be found from this google doc:
https://docs.google.com/document/d/1mtWp8cRl5156VlPWmLpjbJTo5VjosEKNihmmAd8SFeA/edit?usp=sharing)


inst/pages/agglomeration.qmd
```diff
@@ -37,15 +45,15 @@ tse <- GlobalPatterns
 
 ## Agglomerate based on taxonomy rank
 
-One of the main applications of taxonomic information in regards to count data
-is to agglomerate count data on taxonomic levels and track the influence of
-changing conditions through these levels. For this `mia` contains the
+One of the main applications of taxonomic information in sequnencing data
+is to agglomerate counts (such as ASV counts) to specific taxonomic levels 
+and track the influence of changing conditions through these levels. For this `mia` contains the
```
TuomasBorman: Answering your question here.

I think "changing conditions" refers to just sample metadata variables, e.g., diseased or not.

agglomerateByRank() is just agglomerating data so the old version did not make sense.

I think your suggestion is good, can you modify?

inst/pages/agglomeration.qmd
```diff
@@ -164,20 +173,23 @@ altExpNames(tse)
...
+information. Alternatively, you can also use `agglomerateByVariable()`
+to aggregate the data based on `rowData`. In this case the agglomerates 
+are made based on the sample variables. 
 
-For example, a dataset might contain taxa with the same lower-level rank, even
-if their higher-level ranks differ. Thus, `agglomerateByVariable()` should not
-be used for aggregating taxonomy ranks.
+`agglomerateByVariable()` is designed to aggregate data based on other
+criteria than taxa, such as sample groups or clusters. The function operates 
+similarly to feature aggregation, but the `by` parameter must be set to "rows."
```
TuomasBorman: Answering your 2nd question.

agglomerateByRank() operates with rowData, and especially the taxonomy. When it agglomerates the data, it considers the whole taxonomy and ensures that it returrs unique features.

For example, the dataset might include features that have same species-level name but different genus-level name:

- genus x; species a
- genus y; species a

These two features should not be combined as they are different taxa, and `agglomerateByRank()` takes this into account.

However, one might also aggregate data based on other grouping, e.g., butyrate-producers. One cannot do this with `agglomerateByRank()` as this is not related to taxonomy. However, this can be done with `agglomerateByVariable()` which works for both rows or columns.

 `agglomerateByVariable()` can be applied to any variable in rowData (also taxonomy), but it does not take taxonomy into account, it just considers the single column. However, this can lead to equal results if there are no loops in taxonomy, i.e., there are no this kind of taxa:
 
 - genus x; species a
- genus y; species a

but they are like this:

- genus x; species a
- genus y; species b
 
 Check this example:
 
```
 library(mia)
tse <- makeTSE(nrow = 4, ncol = 2)
assayNames(tse) <- "counts"

rowData(tse)[["genus"]] <- c("g1", "g2", "g2", "g3")
rowData(tse)[["species"]] <- c("s1", "s2", "s3", "s3")

tse1 <- agglomerateByRank(tse, rank = "species")
tse2 <- agglomerateByVariable(tse, by = 1L, group = "species")

# agglomerateByRank does not merge anything as these taxonomy are all unique.
# g2;s3 != g3;s3
tse1 |> rowData()
# As agglomerateByVariable considers only species column, it aggregates "s3" group
tse2 |> rowData()

# Add new taxonomy
rowData(tse)[["genus"]] <- c("g1", "g2", "g2", "g2")
rowData(tse)[["species"]] <- c("s1", "s2", "s3", "s3")

tse1 <- agglomerateByRank(tse, rank = "species")
tse2 <- agglomerateByVariable(tse, by = 1L, group = "species")

# Now agglomerateByRank and agglomerateByVariable are working similarly.
tse1 |> rowData()
tse2 |> rowData()
```

However, I think we should clearly state that `agglomerateByRank()` is for taxonomy and `agglomerateByVariable()` for other grouping variables.

I am not sure if it is the most relevant information to tell that these can lead to equal results in certain conditions. Moreover, itcan be hard to understand without examples.

inst/pages/transformation.qmd
```diff
@@ -46,58 +54,65 @@ references.
...
-   instance, in non-parametric statistics.
+   widely used in ecological research, whenever we have reasons to be interested
+   in which taxa are present in which samples more so than their abundances. An
+   example of such case is the process of inferring microbial ecological
+   association networks from covariation of taxa across samples. Here, microbes
+   that are rarely seen in the same samples are expected to have a negative
+   ecological association (either competitive exclusion with each other of
+   different ecological niche preferences). Conversely, looking for negative 
+   correlations in their relative abundances would not be a meaningful measure
+   of  ecological associations.
```
TuomasBorman: I think we are creating more elementary network chapter at some point. That is still our plan.

inst/pages/transformation.qmd
```diff
@@ -46,58 +54,67 @@ references.
...
+   example of such case is the process of inferring microbial ecological
+   association networks from covariation of taxa across samples. Here, microbes
+   that are rarely seen in the same samples are expected to have a negative
+   ecological association (either competitive exclusion with each other of
+   different ecological niche preferences). Conversely, looking for negative 
+   correlations in their relative abundances would not be a meaningful measure
+   of  ecological associations. Moreover, it has been used in machine learning
+   classification models with a good performance [@Giliberti2022,
+   @Karwowska2024].
+
```
TuomasBorman, line 90: I added this back. Hopefully it is now more clear.

inst/pages/transformation.qmd
```diff
@@ -46,58 +54,67 @@ references.
...
-   of bias in the data (see 'rclr' below).
+   it can only be applied to positive values, not zeros. Usual solution for
+   for making values non-zero is to add pseudocount, which adds another type
+   of bias in the data as true taxon absences are not taken into account.
+   This is mitigated by the method of robust clr transformation
+   (see rclr below).
 
  * **hellinger**: Hellinger transformation is equal to the square root of
    relative abundances. This ecological transformation can be useful
    if we are interested in changes in relative abundances.
```
TuomasBorman: Hellinger is not compositionally-aware transformation like CLR. Even though microbiome data is compositional, comparing relative abundances is sometimes enough, especially as ratios are harder to interpret than relative abundances. This sentence reflects to that.

However, this description could be improved. I found an interesting article that could be cited: https://microbiomejournal.biomedcentral.com/articles/10.1186/s40168-023-01747-z

In ML applications, these simpler relative transformations seem to outperform compositional transformations.

TuomasBorman (changes requested):

TuomasBorman:
Any updates on this?

antagomir:
Up @nuorenarra - there are unresolved suggestions, and the PR has got out of sync with the other latest changes ("resolve conflicts")

nuorenarra:
I agree with @TuomasBorman on everything here

antagomir:
There are many unresolved comments. Can you @TuomasBorman and @nuorenarra see how & by whom we should resolve those (or if they can be closed already then just do)?

The conlifcts should be resolved as well, this has got out of sync with the devel branch.

inst/pages/agglomeration.qmd
TuomasBorman: I think this is taken into account.

antagomir (approved):

TuomasBorman (approved):

PR #714, opened 2025-04-06
address issue #697 & expand final section

### LINKED ISSUE #697: ML and transformation
Current, ML chapter applies clr transform to data. However, this paper suggests that one should not apply any transformations:

https://microbiomejournal.biomedcentral.com/articles/10.1186/s40168-023-01747-z

Cite this paper and update the chapter accordingly

Benjamin-Valderrama:
I've added the reference to the suggested in issue #697 and added a discussion about it. In addition, I've expanded the biological interpretation of the plot in the section "Model interpretation".

Ben


TuomasBorman (approved):
Nice! Thanks!

Issue #687, opened 2025-03-16
References to DAA chapter


antagomir:
Good to cite in DAA chapter:

- [Gamboa-Tuz et al. 2025](https://www.biorxiv.org/content/biorxiv/early/2025/02/17/2025.02.13.638109.full.pdf)
- [Pelto et al. 2025](https://arxiv.org/abs/2404.02691)

antagomir:
Elementary methods provide more replicable results in microbial differential abundance analysis  - We benchmarked 14 differential abundance analysis methods in R by examining the consistency and replicability of their results along with their sensitivity. We employed 69 real datasets from 53 microbiome profiling studies (16S & MGS). - Some widely used methods were found to perform inconsistently. - The highest consistency with good sensitivity was obtained with relative abundances or presence/absence with elementary methods (nonparametric test, linear regression/t-test, logistic regression). - Pelto et al. BiB, 2025 https://doi.org/10.1093/bib/bbaf130

antagomir:
lefser is now in Bioconductor and has SummarizedExperiment support - should be cited at least. https://academic.oup.com/bioinformatics/article/40/12/btae707/7908399?login=false

antagomir:
LimROTS
https://bioconductor.org/packages/release/bioc/html/LimROTS.html

@article {Anwar2025,
	author = {Anwar, Ali Mostafa and Jeba, Akewak and Lahti, Leo and Coffey, Eleanor},
	title = {LimROTS: A Hybrid Method Integrating Empirical Bayes and Reproducibility-Optimized Statistics for Robust Analysis of Proteomics Data},
	elocation-id = {2025.02.06.636801},
	year = {2025},
	doi = {10.1101/2025.02.06.636801},
	publisher = {Cold Spring Harbor Laboratory},
[...]

PR #731, opened 2025-05-05
DAA page updated by Juho 

jepelt:
I basically re-wrote the page on differential abundance analysis. It is now based on using simple methods in DAA.

This version still needs re-arranging, simplifying, and polishing of the text and figures, etc..


inst/pages/differential_abundance.qmd
```diff
@@ -127,389 +130,713 @@ table(tse$patient_status, tse$cohort) |>
...
+
+# We use the function plotBoxpplot to illustrate the result for
+# [Ruminococcus]_gauvreauii_group
+plotBoxplot(tse,
+            assay.type = "relabundance",
+            features = "[Ruminococcus]_gauvreauii_group",
+            x = "patient_status",
+            colour.by = "patient_status",
+            fill.by = "patient_status",
+            apply.beeswarm = T)
```
TuomasBorman: I changed the parameter to `point.offset = "quasirandom"`

inst/pages/differential_abundance.qmd
```diff
@@ -127,389 +130,713 @@ table(tse$patient_status, tse$cohort) |>
...
+- `variable` indicates the predictor variable. As `patient_status` is a
+categorical variable, it also tells the group used as the non-reference level
+(here `ADHD` as `Control` was set as the reference level).
+- `estimate` is log(odds ratio). Most importantly, positive values indicate
+abundance being higher in the non-reference group (here `ADHD`). Otherwise,
+estimates can be a bit difficult to interpret for ORMs. 
+
+If we use the standard criterion for statistical significance, namely q-value <
+0.05, we can see that the difference is statistically significant only for the
+genus named as [Ruminococcus]_gauvreauii_group. As the sign of the estimate is
```
TuomasBorman: Instead of hard-coding, could this be dynamic for easier maintenance:

Something like this:

\`r paste0(df[["genus"]], collapse = ", ")\`

inst/pages/differential_abundance.qmd
```diff
@@ -127,389 +130,713 @@ table(tse$patient_status, tse$cohort) |>
...
+# Erysipelatoclostridium by patient_status and cohort
+plotBoxplot(tse,
+            assay.type = "relabundance",
+            features = "Erysipelatoclostridium",
+            x = "patient_status",
+            colour.by = "patient_status",
+            fill.by = "patient_status",
+            facet.by = "cohort",
+            apply.beeswarm = T,
+            add.proportion = T)
```
TuomasBorman: As the visualization is same for all, the visualization methods could be in the end of this chapter.

Visualizations could also include histogram of p-value distribution and including the p-value in the plot?

inst/pages/differential_abundance.qmd
```diff
@@ -127,389 +130,713 @@ table(tse$patient_status, tse$cohort) |>
...
+    .,
+    metadata = colData(tse) |> as.data.frame(),
+    formula = ~ patient_status + cohort)
+  ) |> 
+  bind_rows(.id = "taxon") |> 
+  group_by(variable) |>    # p-values FDR adjusted separately for each variable.    
+  mutate(q_value = p.adjust(p_value, method = "BH")) |> 
+  ungroup()
+
+res_daa_cov |> head() |> kable()
```
TuomasBorman: The updated chapter looks very nice!

However, I am thinking if these custom scripts have the opposite effect: readers/users think they are too complex and just keep on using familiar methods and skip this chapter.

Before we have the package ready, one option is to move these custom scripts to extra material. Here we could have place holders for these and just mention that these scripts can be found from the extra material.

In my opinion, we should consider avoid using custom scripts and use SE-tools whenever possible even though they might be suboptimal in some cases. We can always say their limitations. Pedagogically, these custom scripts might not be optimal.

Proposal:
1. Keep the text more or less as it is. For custom scripts add "the analysis can be found from extra material"
2. Move custom scripts to another file that will be listed as "Extra material"
3. In this chapter, use Masslin3 as it is supporting SE
4. When we have DAA package ready, we can easily update this chapter

TuomasBorman (changes requested):
Looks nice! See comments

inst/pages/differential_abundance.qmd
jepelt: Ok, good. I'll use it also when updating the visualization section.

inst/pages/differential_abundance.qmd
jepelt: Hmm, I think hard coding is fine here. If we use e.g. a different dataset later, the text will likely need some (other) modifications anyway. So, I would not just automatically take the first taxon on the list and write that in the text.

inst/pages/differential_abundance.qmd
jepelt: Yes, I will move the visualizations at the end.

p-value in the plot is a good idea.

Hmm, I'm not sure what you mean by a histogram of p-values.. Is that a standard figure? I would say a figure with estimates and confidence intervals would be better. (Like the one that is there now in OMA book for MaAsLin3.)

inst/pages/differential_abundance.qmd
jepelt: Thanks for the comments! I agree on almost everything.

My only real concern is in using maaslin3 in the "normal" DAA. So, maaslin3 is good in analyzing positive abundances. In analyzing presence/absence it is not very good but, ok, it is acceptable for me to use it in that, too, if the limitations are mentioned (and custom scripts are given in the extra material). However, maaslin3 does not actually perform single DAA test. You can only get combined p-values from P/A test and positive abundance test (and p-values are combined in a bit conservative way). Consequently, you also don't get one but two estimates for each taxon. Using maaslin3 would also be a bit against to what I am suggesting in the "introduction"..

So, could the normal DAA performed with e.g. MaAsLin2 package? It has a few annoying things but it is easy to use and performs DAA in a way that I am suggesting in the introduction. The only difference in its use (vs. maaslin3) is that metadata and counts should be extracted:

Maaslin2(input_data = assay(tse), input_metadata = colData(tse) |>  as.data.frame(), ...)
vs.
maaslin3(tse, ...)

inst/pages/differential_abundance.qmd
TuomasBorman: btw, adding p-values to boxplot would benefit the users more broadly. Moreover, adding them is not trivial task (especially with rather complex visualization that can be done with plotBoxplot()). I could check if this can be incorporated to the function itself.

With histogram I was referring this kind of plot 
image

Not sure if it is commonly used.

We could keep the MaAsLin3 figure until we have the DAA package ready (good to have multiple different ways to visualize the results). We can then create our own visualizations specifically for the output of its functions.

inst/pages/differential_abundance.qmd
TuomasBorman: In my opinion, it is ok to use Maaslin2 (but we could use the Maaslin3 for positive abundances if it fits to the chapter). I think one of the most important point of OMA is to explain openly and clearly why something is done. Some people might disagree but that hopefully leads to discussion and overall advances of field. Nothing in the book is immune to  discussion, and anything can be changed if someone provides better evidence.

inst/pages/differential_abundance.qmd
antagomir: Is there some downside of removing hard-coding? It seems a straightfwd thing to add.

inst/pages/differential_abundance.qmd
antagomir: Summary plots on the overall DAA results could be useful to have.

I am not sure if they are something to have with individual boxplots since they are related to DAA on the entire data set. Summarizing some options:

1. P-value histogram: sometimes checked but not widely used in reporting in my experience.
2. Volcano plot (p-value vs. effect size scatterplot): more standard way to visualize outcomes from this kind of multiple testing, I would consider adding that instead. 
3. Figure with estimates and confidence intervals is rather standard and would be nice to have indeed. 

I would be most in favor or 2-3 but it is fine to add them later with DAA package.

inst/pages/differential_abundance.qmd
antagomir: Im ok with either, especially if our own DAA pkg is proceeding soon anyway.

antagomir:
One thing I am wondering is - there were substantial contributions from a few people to the DAA chapter over the years, at least HE, MC, LY, and others and including example workflows on several alternative, specific methods (ANCOMBC, MaaslinX, DESeq, LinDA etc etc). In some cases these examples might still become handy, for instance in review situations. 

Should we include the past workflows in supplementary material or github repo that is hidden from the TOC but linked from DAA chapter in some suitable way as extra material? And that would not require active updates, it would be more like a historical reference on how certain common methods could be run shall there be a particular need.

jepelt:
I updated the page again.
- "Normal DAA" is now based on MaAsLin2
- Presence/absence analysis and DAA for positive abundances are based on MaAsLin3 (Maybe in the P/A analysis the library size should be controlled for..) 
- ORM and Firth logistic regression are now in extra materials (they are now at the end of the document but will be moved to the Extra material chapter.)
- Visualizations are now at the end of the document.
- The text still needs some polishing etc.
- I could not answer to Leo's comment on the older ANCOM-BC etc things. But yeah, maybe they can be put somewhere. I have to check, however, how they work with the latest versions of the packages etc.

TuomasBorman:
> One thing I am wondering is - there were substantial contributions from a few people to the DAA chapter over the years, at least HE, MC, LY, and others and including example workflows on several alternative, specific methods (ANCOMBC, MaaslinX, DESeq, LinDA etc etc). In some cases these examples might still become handy, for instance in review situations.
> 
> Should we include the past workflows in supplementary material or github repo that is hidden from the TOC but linked from DAA chapter in some suitable way as extra material? And that would not require active updates, it would be more like a historical reference on how certain common methods could be run shall there be a particular need.

If there is need for this kind of reference (and git history does not fill it sufficiently) we can have these in extra material.

However, I am little bit concerned about having material in the book that is not updated and recommended or otherwise not follow the same quality as the rest of the book. It easily becomes outdated and any case we are responsible on that extra material also.

This problem more or less disappears when we start to use the versioning and follow the Bioconductor release cycle.

inst/pages/differential_abundance.qmd
jepelt: Maybe there isn't any. So, I'll remove it in the next update.

inst/pages/differential_abundance.qmd
jepelt: Instructions to make a figure with estimates and CIs is now added in the visualizatin section.

antagomir:
Was this closed  on purpose before merging it to devel branch @jepelt ?

jepelt:
No, I just somehow accidentally closed this! Can this be "unclosed"? Or can I just create a new pull request?

antagomir:
There is this button Reopen PR below.

antagomir:
I just pressed it. Now the PR is open again..

jepelt:
Ok, thanks!

jepelt:
I made now some updates:
- The extra materials are now moved into the extra materials chapter.
- Sequencing depth is now controlled for in presence/absence analysis as suggested by the authors of maaslin3
- Literature review section is slightly updated
- Some typos corrected, etc
- The hard coding of genus names is still there, however. I decided to leave it like that as I don't always refer in the text to the genera with, for instance, the smallest p-value(s). MaAsLin2 also has difficulties with some special characters (`[`, `]`) in the genera names.

inst/pages/differential_abundance.qmd
```diff
@@ -1,120 +1,137 @@
...
+
+## Performing differential abundance and presence/absence analysis
+
+In this section we show how to perform differential abundance or
+presence/absence analysis as recommended above. We illustrate the basic case of
+comparing abundances or presence/absences of taxa between two experimental
+groups. Furthermore, we illustrate how to add covariates in the analysis and
+how to examine the interaction effect of two variables.
+
+We use the Tengler2020 dataset from mia package. The dataset includes samples
```
antagomir: It is "Tengeler2020" not "Tengler2020"

antagomir (commented):
One main comment: this PR recommends to use non-parametric test for relative abundances (e.g. TSS normalization). This would ignore the effect of compositionality, which has been demonstrated to lead to problems also in DA analysis (see e.g. the classic Gloor et al. 2017 paper: https://www.frontiersin.org/journals/microbiology/articles/10.3389/fmicb.2017.02224/full). What is our main argument and evidence that the compositionality can be safely ignored in general?

jepelt:
> One main comment: this PR recommends to use non-parametric test for relative abundances (e.g. TSS normalization). This would ignore the effect of compositionality, which has been demonstrated to lead to problems also in DA analysis (see e.g. the classic Gloor et al. 2017 paper: https://www.frontiersin.org/journals/microbiology/articles/10.3389/fmicb.2017.02224/full). What is our main argument and evidence that the compositionality can be safely ignored in general?

Basically this pre-print by Gamboa-Tuz, et al. (2025) https://www.biorxiv.org/content/10.1101/2025.02.13.638109v1.full

So, yes, MB data are compositional (and there is also the taxon-wise bias https://pmc.ncbi.nlm.nih.gov/articles/PMC6739870/ ). The question is, however, if/how these bias(es) can be corrected. "Compositional methods" (e.g. CLR based methods) work very nicely on simulated data but on the real data in the above Gamboa-Tuz paper they were among the worst methods (while TSS performed well).

Compositionality has also been studied on real data here https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1010284 but TSS was not included there, and they also did not seem to give any clear answer which of the methods (ALDEx2, DESeq2, edgeR, ANCOM-BC, scran) berformed better. I also studied this quickly in an additional analysis using one of their datasets (Barlow mice study) and found that TSS methods performed at least comparably to other methods (Figure A15 in Appendix https://academic.oup.com/bib/article/26/2/bbaf130/8093585 ). But my analysis was very limited.

So, it is still probably rather open question how compostionality should be treated (or can it be treated), but the (very scarce!) real data evidence seems to suggest that, in real life, compositional methods may not treat compositionality any better than TSS. Therefore, I think it is probably better to keep things simple..

But yes, this should be discussed in the text more explicitly! Now I only have the Gamboa-Tuz paper as one of the bullet points in the section "Recommendations based on recent literature", and compositionality is not even explicitly mentioned there. Maybe compositionality should even have its own subsection somewhere there?

I now recommend linear regression with TSS (MaAsLin2) as the main option (non-parametric method is only an alternative in extra materials) but maybe I should give compositional alternative, too? That could be most easily done by using MaAsLin2 with CLR normalization/transformation (or some other non-TSS normalization). Then the nornalization would be the only thing that would change, and the results between TSS and CLR can be directly compared. (If I recommended completely different methods e.g. ALDEx2 or ANCOM-BC(2) there would be also other things that change, and then the effect of compositionality would be difficult to evaluate..)

(By the way MaAsLin3's positive abundance analysis has compositionality adjustment as a (default) option so that would be easy to recommend as an alternative.)

inst/pages/differential_abundance.qmd
jepelt: > It is "Tengeler2020" not "Tengler2020"

Ok, thanks, I'll correct this.

jepelt: > One main comment: this PR recommends to use non-parametric test for relative abundances (e.g. TSS normalization). This would ignore the effect of compositionality, which has been demonstrated to lead to problems also in DA analysis (see e.g. the classic Gloor et al. 2017 paper: https://www.frontiersin.org/journals/microbiology/articles/10.3389/fmicb.2017.02224/full). What is our main argument and evidence that the compositionality can be safely ignored in general?

Basically this pre-print by Gamboa-Tuz, et al. (2025) https://www.biorxiv.org/content/10.1101/2025.02.13.638109v1.full

So, yes, MB data are compositional (and there is also the taxon-wise bias https://pmc.ncbi.nlm.nih.gov/articles/PMC6739870/ ). The question is, however, if/how these bias(es) can be corrected. "Compositional methods" (e.g. CLR based methods) work very nicely on simulated data but on the real data in the above Gamboa-Tuz paper they were among the worst methods (while TSS performed well).

Compositionality has also been studied on real data here https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1010284 but TSS was not included there, and they also did not seem to give any clear answer which of the methods (ALDEx2, DESeq2, edgeR, ANCOM-BC, scran) berformed better. I also studied this quickly in an additional analysis using one of their datasets (Barlow mice study) and found that TSS methods performed at least comparably to other methods (Figure A15 in Appendix https://academic.oup.com/bib/article/26/2/bbaf130/8093585 ). But my analysis was very limited.

So, it is still probably rather open question how compostionality should be treated (or can it be treated), but the (very scarce!) real data evidence seems to suggest that, in real life, compositional methods may not treat compositionality any better than TSS. Therefore, I think it is probably better to keep things simple..

But yes, this should be discussed in the text more explicitly! Now I only have the Gamboa-Tuz paper as one of the bullet points in the section "Recommendations based on recent literature", and compositionality is not even explicitly mentioned there. Maybe compositionality should even have its own subsection somewhere there?

I now recommend linear regression with TSS (MaAsLin2) as the main option (non-parametric method is only an alternative in extra materials) but maybe I should give compositional alternative, too? That could be most easily done by using MaAsLin2 with CLR normalization/transformation (or some other non-TSS normalization). Then the nornalization would be the only thing that would change, and the results between TSS and CLR can be directly compared. (If I recommended completely different methods e.g. ALDEx2 or ANCOM-BC(2) there would be also other things that change, and then the effect of compositionality would be difficult to evaluate..)

(By the way MaAsLin3's positive abundance analysis has compositionality adjustment as a (default) option so that would be easy to recommend as an alternative.)

antagomir: Ok beautiful. 

Gamboa-Tuz et al. (2025), your paper plus some other papers and explanation could do as a justification.

Ideally, we would have brief guidelines if recommendations differ by situation (e.g. low vs high diversity; low vs. high number of samples; low vs. high number of features etc). To be checked in more detail. 

Keep in mind that OMA is not intended to replicate material that can be easily accessed elsewhere. Descriptions could be short and cite the additional papers where more details can be found.

The [MicrobiomeBenchmarkData](https://www.bioconductor.org/packages/release/data/experiment/html/MicrobiomeBenchmarkData.html) package with biologically relevant benchmarking cases is readily available as TreeSE; should we use this rather than the Tengeler data set (which was originally intended for a different purpose in an applied study on gut-brain analyses)?

Also future benchmarks could look more into this.

antagomir:
> > One thing I am wondering is - there were substantial contributions from a few people to the DAA chapter over the years, at least HE, MC, LY, and others and including example workflows on several alternative, specific methods (ANCOMBC, MaaslinX, DESeq, LinDA etc etc). In some cases these examples might still become handy, for instance in review situations.
> > Should we include the past workflows in supplementary material or github repo that is hidden from the TOC but linked from DAA chapter in some suitable way as extra material? And that would not require active updates, it would be more like a historical reference on how certain common methods could be run shall there be a particular need.
> 
> If there is need for this kind of reference (and git history does not fill it sufficiently) we can have these in extra material.
> 
> However, I am little bit concerned about having material in the book that is not updated and recommended or otherwise not follow the same quality as the rest of the book. It easily becomes outdated and any case we are responsible on that extra material also.
> 
> This problem more or less disappears when we start to use the versioning and follow the Bioconductor release cycle.

This tutorial may become rather defining source for some practitioners of microbiome data analysis. 

The different methods may have different pros & cons depending on the exact situation, and the authors may have done remarkable extra effort to support the Bioc/TreeSE framework.

If/when we only recommend one or very few approaches, it may unfairly advertise some methods over others as the "_de facto_" standard of doing DAA (or some other analysis). On the other hand, having many different methods would complicate the analyses unnecessarily. 

The primary focus should be on concepts, on the differerent ways to think about and execute DAA. When it is not feasible to demonstrate all methods, citations and a link to the external vignettes can be provided in many cases.

antagomir:
See also #687

jepelt:
I (again) closed this accidentally on Friday but re-opened this now.. Anyway, I updated the DAA page again. (Can you see the updates..?)

Based on our discussions, the page is now shorter and the visualizations are moved to the context of DAA. I also added a short section "Alternative options" at the end, where I very briefly say something about more advenced compositionality adjustments.

antagomir:
Yes it shows

antagomir (approved):

signynas:
Hi!

While going through the DAA chapter, I noticed it seems like the ZicoSeq section at the end of the chapter ends kind of abruptly. Below the complete section:

```
Subsequently, we demonstrate DAA with ZicoSeq, a method based on linear models and permutation. Further details can be found in this tutorial. This approach has been assessed to exhibit high power and a low false discovery rate, which has the following components:

1. Winsorization to decrease the influence of outliers;

2. Posterior sampling based on a beta mixture prior to address sampling variability and zero inflation;

3. Reference-based multiple-stage normalization to address compositional effects.
```

Specifically the sentence "Further details can be found in this tutorial. " seems misleading as it is not elaborated on at all in that chapter. Was it left in mistakenly or is it unfinished?

I also removed the words "Additional resources;" that was left most likely by accident on the last bullet point. The section after it starts with the words Additional resources.

Additionally, in most of the other chapters in the book, R packages and functions are formatted with `<package>`; however, in this chapter they are mostly just plain text, except for a few italized ones. Should they be consistent with the other chapters, or have they specifically not been formatted with the `` since they are referred to as methods (e.g., ALDEx2, ANCOM-BC, DESeq, etc.)?

jepelt:
> Hi!
> 
> While going through the DAA chapter, I noticed it seems like the ZicoSeq section at the end of the chapter ends kind of abruptly. Below the complete section:
> 
> ```
> Subsequently, we demonstrate DAA with ZicoSeq, a method based on linear models and permutation. Further details can be found in this tutorial. This approach has been assessed to exhibit high power and a low false discovery rate, which has the following components:
> 
> 1. Winsorization to decrease the influence of outliers;
> 
> 2. Posterior sampling based on a beta mixture prior to address sampling variability and zero inflation;
> 
> 3. Reference-based multiple-stage normalization to address compositional effects.
> ```

Hi, are you reviewing the current version of the DAA page, and not the version in this pull request? This pull request/updated version should not include such paragraph about ZicoSeq.

signynas:
> 
> Hi, are you reviewing the current version of the DAA page, and not the version in this pull request? This pull request/updated version should not include such paragraph about ZicoSeq.

Ahh that's right, I am reviewing the current version, my bad.

TuomasBorman (approved):
Thanks!

PR #749, opened 2025-06-10
Update support info, remove topGO

TuomasBorman:


TuomasBorman:
topGO package fails to install, and this has been lasting long. It seems that it is not maintained anymore. Original author has made the GitHub repo private: https://github.com/llrs/topGO/issues/4#issuecomment-2079033966

PR #752, opened 2025-06-12
Cross association function: fix deprecated arguments

artur-sannikov:
Some arguments are deprecated in `getCrossAssociation` function. I fixed it before it becomes a problem.


TuomasBorman (approved):
Thanks!

PR #757, opened 2025-07-01
Update alpha diversity example

mm-hak:
The example was modified to use the default indices returned by addAlpha():
dbp_dominance, faith_diversity, observed_richness, and shannon_diversity.  
All subsequent code chunks were updated so that the referenced index names match this output.

### DIFF
--- a/inst/pages/alpha_diversity.qmd
+++ b/inst/pages/alpha_diversity.qmd
@@ -135,8 +135,8 @@ library(mia)
 data("GlobalPatterns")
 tse <- GlobalPatterns
 
-# The 'index' parameter allows computing multiple diversity indices simultaneously.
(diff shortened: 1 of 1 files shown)


TuomasBorman (approved):
Thanks!

antagomir:
This PR could also include citation to the publication that recommended that complementary set of indices (as described in the addAlpha manpage)?

TuomasBorman:
It is already included in line 97 (Book has not rendered in a while, there has been errors one after another. At some point, we have to discuss how to manage the book with all the dependencies)

antagomir:
Perfect.

antagomir:
Getting the entire book rendered would be a high priority I guess.

PR #762, opened 2025-07-03
Fixed formatting issues and error in example solutions

raivo-otus:
Added .csv into the .gitignore.

Fixed formatting issues that where not being parsed corrrectly. Should be inline with Quarto styling now.

Fixed typo and code in example_solutions.qmd

### DIFF
--- a/inst/pages/example_solutions.qmd
+++ b/inst/pages/example_solutions.qmd
@@ -319,16 +319,16 @@ p1 + p2
 
 # Subset the abundance table
 clr_matrix <- assay(tse, "clr")
-clr_matrix[1:10, 1:5]
+clr_matrix[1:100, 1:10]
 
-# Agglomerate data by Phylum rank
-tse_phylum <- agglomerateByRank(tse, rank = "Phylum")
+# Agglomerate data by ranks
+tse <- agglomerateByRanks(tse)
 
-# Apply transformation to Phylum data
-tse_phylum <- transformAssay(
-    tse_phylum, assay.type = "counts", method = "relabundance")
-tse_phylum <- transformAssay(
-    tse_phylum, assay.type = "counts", method = "clr", pseudocount = 1)
+# Apply transformations across altExp
+tse <- transformAssay(
+    tse, altexp = altExpNames(tse), assay.type = "counts", method = "relabundance")
+tse <- transformAssay(
+    tse, altexp = altExpNames(tse), assay.type = "counts", method = "clr", pseudocount = 1)
 
 # phiLR transformation
 tse <- transformAssay(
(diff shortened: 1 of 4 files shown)


inst/pages/example_solutions.qmd
```diff
@@ -322,14 +322,13 @@ clr_matrix <- assay(tse, "clr")
...
 # Apply transformation to Phylum data
-tse_phylum <- altExp(tse, "Phylum")
-tse_phylum <- transformAssay(
-    tse_phylum, assay.type = "counts", method = "relabundance")
-tse_phylum <- transformAssay(
-    tse_phylum, assay.type = "counts", method = "clr", pseudocount = 1)
+altExp(tse, "Phylum") <- transformAssay(
+    altExp(tse, "Phylum"), assay.type = "counts", method = "relabundance")
+altExp(tse, "Phylum") <- transformAssay(
+    altExp(tse, "Phylum"), assay.type = "counts", method = "clr", pseudocount = 1)
```
TuomasBorman, line 332: This is indeed good practice. However, I tend to think that these example solutions should be as simple as possible. What do you think?

TuomasBorman (commented):
Thanks! Check comment

TuomasBorman:
One quick note:

The book was not failing because of these. It is currently failing because of package mismatches. There was major changes in SummarizedExperiment, and it is still causing as headache (probably Bioconductor Docker does not have the updated, fixed package versions...).

Related issue: https://github.com/drisso/SingleCellExperiment/issues/83

With minimal changes, you could add this to DESCRIPTION's remotes. It should then install SE from source which should work. We can then remove the line in the future:

Bioconductor/SummarizedExperiment

inst/pages/example_solutions.qmd
raivo-otus: The exercise for which this solution is was worded as;

- Agglomerate the data with [agglomerateByRanks()](https://rdrr.io/pkg/mia/man/agglomerate-methods.html) Transform data now with altexp = altExpNames(tse).

The Exercise question may need some rewording, but I think the solution is in line with the current question. Do you have a simpler solution in mind?

raivo-otus:
> One quick note:
> 
> The book was not failing because of these. It is currently failing because of package mismatches. There was major changes in SummarizedExperiment, and it is still causing as headache (probably Bioconductor Docker does not have the updated, fixed package versions...).
> 
> Related issue: [drisso/SingleCellExperiment#83](https://github.com/drisso/SingleCellExperiment/issues/83)
> 
> With minimal changes, you could add this to DESCRIPTION's remotes. It should then install SE from source which should work. We can then remove the line in the future:
> 
> Bioconductor/SummarizedExperiment

Thanks for the note. Testing this solution currently. Will commit a separate PR if succesful.

raivo-otus:
Should be a bit more clear now, and the solution is as intended I believe. 

Fixed the numbering of tasks as well.

TuomasBorman (approved):
Thank you!

PR #777, opened 2025-08-04
Fix/render error

raivo-otus:
Explicitly call function in agglomeration.qmd to prevent function not found error when rendering.

Set problematic chunks in extra_materials.qmd to `eval: false` as a band-aid for obscure bug when rendering.

Remove reduntant chunk options in import.qmd.

### DIFF
--- a/inst/pages/agglomeration.qmd
+++ b/inst/pages/agglomeration.qmd
@@ -1,6 +1,10 @@
 # Agglomeration {#sec-agglomeration}
 
-```{r setup, echo=FALSE, results="asis"}
+```{r}
+#| label: setup
+#| echo: false
+#| results: asis
+
 library(rebook)
 chapterPreamble()
 ```
--- a/inst/pages/agglomeration.qmd
+++ b/inst/pages/agglomeration.qmd
@@ -86,7 +86,7 @@ Phylum rank.
 
 Since we specified `update.tree = TRUE`, the phylogenetic tree has also been
 agglomerated. This is evident from the tree, which now has only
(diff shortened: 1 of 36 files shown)


inst/pages/agglomeration.qmd
```diff
@@ -86,7 +86,7 @@ Phylum rank.
 
 Since we specified `update.tree = TRUE`, the phylogenetic tree has also been
 agglomerated. This is evident from the tree, which now has only
-`r length(rowTree(tse_phylum)$tip.label)` tips, each corresponding to a single
+`r length(TreeSummarizedExperiment::rowTree(tse_phylum)$tip.label)` tips, each corresponding to a single
```
TuomasBorman: What is the motivation?

TuomasBorman (changes requested):
Thanks!

Can you instead set the eval: false globally, so we do not need to control these multiple chunks separately

TuomasBorman:
> Thanks!
> 
> Can you instead set the eval: false globally, so we do not need to control these multiple chunks separately


This might work


knitr::opts_chunk$set(
  eval = FALSE
)

inst/pages/agglomeration.qmd
TuomasBorman: The function should come with mia as TreeSE is direct dependency, I think

TuomasBorman: And mia is called above

raivo-otus: I was getting a function not found error when rendering, unless the function called inline was called explicitly as so.

raivo-otus: Seems to be an issue on my end. Works here.

TuomasBorman (approved):

Issue #780, opened 2025-08-07
Enforce certain code styling across the book


artur-sannikov:
# Problem

The code style at the moment is rather inconsistent. Example:

In Differential abundance [chapter](https://microbiome.github.io/OMA/docs/devel/pages/differential_abundance.html). See comments about parentheses below.

```R
ancombc2_out <- [ancombc2](https://rdrr.io/pkg/ANCOMBC/man/ancombc2.html)(
    data = tse,
    assay.type = "counts",
    fix_formula = "patient_status",
    p_adj_method = "fdr",
    prv_cut = 0,
    group = "patient_status",
    struc_zero = TRUE,
    neg_lb = TRUE,
    # multi group comparison is deactivated automatically
    global = TRUE) # The parenthesis is where the code ends
```

But below:

```R
maaslin3_out <- [maaslin3](https://rdrr.io/pkg/maaslin3/man/maaslin3.html)(
    input_data = tse,
    output = "DAA_example",
    formula = "~ patient_status",
    normalization = "TSS",
    transform = "LOG",
    verbosity = "ERROR"
    ) # But here it's on the next line. I think it should be indented 
```
# Proposed solution
There are packages that deal with this inconsistency automatically, for example, [styler](https://styler.r-lib.org). We can always run it before we push the code using something like [precommit](https://github.com/lorenzwalthert/precommit). However, with a big number of contributors, it's challenging to enforce.

I think a better solution is to use a [GitHub action](https://styler.r-lib.org/articles/third-party-integrations.html) to automatically style the code directly on GitHub.

I can have a look into it if the maintainers think it's a good idea.

TuomasBorman:
Good idea. In fact, this has been our plan, but this seems better approach.. It should be run only in PRs. 

E: It would be nice, if you can create a PR. The styling must follow Bioconductor standards. For instance, 4 space indentation, and max. 80 character wide lines. We can check this in detail in PR.

antagomir:
Check with jpelto as well, not sure what is the current plan whether ANCOMBC2 will be kept as an executed example.

artur-sannikov:
@antagomir. It's ok if it's there. `styler` works on any code. We can remove ANCOMBC2 if it's not required, it will not change anything.

@TuomasBorman, `styler` only works with code, not text. I already see that some text exceeds 80 character limit. Is it critical?

TuomasBorman:
No, it is just good practice

PR #783, opened 2025-08-07
Add code styling to GHA

artur-sannikov:
See #780.

It's setup to have indentation of 4 spaces according to Bioconductor [guidelines](https://contributions.bioconductor.org/r-code.html#coding-style).

One point that is difficult to implement is 

* No space around “=” when using named arguments to functions. This: somefunc(a=1, b=2)

In `mia` this convention is [not followed strictly](https://github.com/microbiome/mia/blob/devel/R/runCCA.R#L356), so this current implementation should work?


.github/workflows/style.yaml
```diff
@@ -0,0 +1,73 @@
+on:
+  push:
+    branches:
+      - devel
+      - RELEASE_**
```
TuomasBorman: This seems to run when pushing to devel branch (without PR)?

Can this cause any issues? I see it little bit concerning if GHA can modify our "main" (in this case branch called devel) branch. Everything that is merged into "main" should be accepted by us, and GHA should not be able to modify our main code base. Moreover, this can cause problems in the future if the GHA is modified in the future, and this issue is not recognized.

I would be more comfortable, if GHA would have access only to PRs (anyways everything should go through it).

TuomasBorman (changes requested):
Looks awesome! See the comment.

TuomasBorman:
> See #780.
> 
> It's setup to have indentation of 4 spaces according to Bioconductor [guidelines](https://contributions.bioconductor.org/r-code.html#coding-style).
> 
> One point that is difficult to implement is
> 
>     * No space around “=” when using named arguments to functions. This: somefunc(a=1, b=2)
> 
> 
> In `mia` this convention is [not followed strictly](https://github.com/microbiome/mia/blob/devel/R/runCCA.R#L356), so this current implementation should work?

It is fine. The main thing is to harmonize the style, and I think current style looks good

.github/workflows/style.yaml
artur-sannikov: Should work now.

TuomasBorman (approved):
Thanks!

artur-sannikov:
@TuomasBorman, FYI, this workflow can be also used in miaverse packages.

Issue #769, opened 2025-07-15
Inconsistent filenames in Docker image in relation to OMA book


signynas:
When looking for improvements in the OMA book and editing them in the RStudio Docker image, I noticed that the filenames of the chapters are not consistent with the chapter names in the book. This makes finding the correct files to add improvements to much harder, as I have to check many potential files to then see the heading and know if it is the corresponding chapter file or not. I would suggest to rename the files (that are not already consistent with the book) in the docker image, so they correspond with the chapter names. An example of this is chapter 25 - Multiomics prediction and classification, that has the filename integrated_learner in the image. There are multiple other files like these, some more transparent than others. For simplicity, even though the names have similar meaning, it would be better for the names to match.

Would this be an OK change for me to make and then create a PR of it?

antagomir:
Ok to me and sounds useful.

One potential issue is the risk of conflicts as there are currently some open PRs. This could be started with sections that no one seems to be actively working on?

signynas:
Sounds good. Is there a way to do some general poll on which sections people are currently working on?

antagomir:
You could ask in our Mattermost but I am not sure how reliable that method is. If you change this in one batch for those files where it seems "safe" and then the rest one-by-one as separate PRs (can be more easily cancelled then)?

artur-sannikov:
Furthermore, https://github.com/microbiome/OMA/pkgs/container/oma has old images. I assumed the [workflow](https://github.com/microbiome/OMA/blob/devel/.github/workflows/rworkflows.yml) builds them automatically.

PR #785, opened 2025-08-11
Rename file for chapter 31 - Contributions

signynas:
Discussed in issue #769 .

I renamed the file for Chapter 31 - Contributions. It was previously named acknowledgments.qmd, and is now contributions.qmd, to be consistent with the book chapter name. I also made changes in the _book.yml files, by correcting the chapter name there so it renders correctly.


TuomasBorman (approved):
thanks!

PR #797, opened 2025-09-07
Update DESCRIPTION

TuomasBorman:

PR #799, opened 2025-09-11
Fix styler gha "fatal: You are not currently on a branch."

artur-sannikov:
There was an error `fatal: You are not currently on a branch`. For example, [here](https://github.com/microbiome/OMA/actions/runs/17555593280/job/49858807895). 

I think this should fix it  (see https://github.com/actions/checkout/issues/317#issuecomment-1221335371)

```yaml
uses: actions/checkout@v3
with:
  ref: ${{ github.ref_name }}
```

At least, it  work in my [Action](https://github.com/artur-sannikov/OMA/actions/runs/17635477585/job/50110767188).

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -19,6 +19,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          ref: ${{ github.ref_name }}
       - name: Setup R
         uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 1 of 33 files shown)


TuomasBorman (changes requested):
Thanks, it was also fixed here but then I realized that better to wait other PRs to be merged. There is one bigger one.

https://github.com/microbiome/OMA/pull/797

TuomasBorman:
https://github.com/microbiome/OMA/pull/802

PR #812, opened 2025-09-22
Removed deprecated package reference from miaverse.qmd

signynas:
In the miaverse chapter, I removed a reference to the package [CBEA](https://www.bioconductor.org/packages/release/bioc/html/CBEA.html), because it has been deprecated.


TuomasBorman (approved):
Great!

PR #821, opened 2025-11-01
Move part introductions chapters to part prefaces

RiboRings:
Streamlined prefaces to parts so they are not separate numbered chapters. Based on this docs: https://quarto.org/docs/books/book-structure.html#parts-appendices

### DIFF
--- a/inst/assets/_book.yml
+++ b/inst/assets/_book.yml
@@ -34,9 +34,8 @@ book:
       chapters:
         - pages/taxonomy.qmd
         - pages/wrangling.qmd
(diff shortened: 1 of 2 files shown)


TuomasBorman (commented):
Seems good! Thanks!

Can you check that we have either "multiomics" or "multi-omics"

We use multi-omics in our manuscript so that might be the choice?

RiboRings:
Ok! I changed the spelling to multi-omics.

TuomasBorman (approved):
Thanks!

PR #830, opened 2026-01-15
Refactor/switch to american english

### LINKED ISSUE #800: Use US English
Ensure that the book is using US English. Bioconductor documentation and books are written in US English so it is good to harmonize.

@signynas

artur-sannikov:
See #800.

I think I covered all of British spelling.


TuomasBorman (approved):
Awesome.

TuomasBorman:
Btw, what script did you use? Thinking about the future use...

What do you think, should this be automated like "Style code"?

artur-sannikov:
I just used https://www.npmjs.com/package/american-british-english-translator to look see what spelling needs to be changed, and then just used something like for each word:

```
find -name "*.qmd" -or -name "*.Rmd" | xargs sed -i 's/minimises/minimizes/g'
```

We can use a bash script to do this automatically at every PR.

artur-sannikov:
Don't forget to fix the permissions for styler: https://github.com/microbiome/OMA/actions/runs/21026866334/job/60453346460

TuomasBorman:
> Don't forget to fix the permissions for styler: https://github.com/microbiome/OMA/actions/runs/21026866334/job/60453346460

It is working for OMA branches, but the styler does not have access to forks. It might be possible somehow to add them as " Maintainers are allowed to edit this pull request. ", but don't know how. Can you check?

I.e., can styler have "maintainer" status?

TuomasBorman:
> I just used https://www.npmjs.com/package/american-british-english-translator to look see what spelling needs to be changed, and then just used something like for each word:
> 
> ```
> find -name "*.qmd" -or -name "*.Rmd" | xargs sed -i 's/minimises/minimizes/g'
> ```
> 
> We can use a bash script to do this automatically at every PR.

Can you make one? Create new GHA for that so we can remove it if it is not working

artur-sannikov:
Note: some words do not have to be substituted, such as `colour` in the function parameters.

artur-sannikov:
Should be fine now. Let me know, and I wish squash the commits.

TuomasBorman:
It seems that the GHA is failing to push. "Style" GHA is also pushing --> can you check that these both are using same approach

artur-sannikov:
That's expected because the `branch/switch-to-american-english` does not exist in the repo. On my fork, it works fine: https://github.com/artur-sannikov/OMA/actions/runs/21242510916/job/61123638447

But I will see what I can do to fix it.

artur-sannikov:
Now the conversion GHA is fixed. The style fails because of the 403 error (access forbidden). I'm not entirely sure what's going on here because I gave it the write permissions to the repo.

However, I plan to switch to [this Action](https://github.com/stefanzweifel/git-auto-commit-action) for commit and push the changes. Right now, it's my [custom script](https://github.com/microbiome/OMA/blob/devel/.github/workflows/style.yaml#L66).

.github/workflows/american-english.yaml
```diff
@@ -0,0 +1,52 @@
...
+          fetch-depth: 0
+      - name: Convert from American English to British English
+        run: |
+          find . -type f -iname "*.r" -or -iname "*.[qr]md" \
+                 -or -iname "*.rmarkdown" |
+            xargs sed -i \
+            -e 's/visualise/visualize/' \
+            -e 's/visualisation/visualization/' \
+            -e 's/summarise /summarize /' \
+            -e 's/summarised /summarized /' \
```
TuomasBorman: Why there is space after summarised? Does it mean that this does not work for

> This sentence is summarised.

TuomasBorman (approved):

TuomasBorman:
Thanks, looks good! Check the last comment.

.github/workflows/american-english.yaml
artur-sannikov: > Why there is space after summarised? Does it mean that this does not work for
> 
> This sentence is summarised.

The reason is that we have a `summarise` function, so `summarise()` would be
matched. That's unwanted behavior. I would use the negated set here
(https://www.regular-expressions.info/charclass.html).

I tested it briefly: it will not work with `summarise` (no space character),
but I suppose this situation should not occur.

So, something like this: "s/\<summarise\>([^(])/summarize\1/" and
"s/\<summarised\\>([^(])/summarized\1/". Not sure if this can be merged into
one rule. Need some Regex gurus here.

TuomasBorman: For now, can you move the GHA related stuff to the PR: https://github.com/microbiome/OMA/pull/840

The text is good to go but this might need more thinking?

TuomasBorman: I still feel this is not safe enough as it can match with function names or variables and lead to problems in the future. Maybe words inside code chunks should be ignored, but I would already merge the qmd files from this PR to get forward.

@artur-sannikov

artur-sannikov: We can use the negation pattern `^` to ignore some blocks, but I feel like
we are entering the bash script area here. Do you know someone with better 
regex knowledge?

TuomasBorman: You might be the best candidate that I know. Also, this might need more sophisticated approach (for instance dedicated R function).

As the PR is good otherwise, I propose that the GHA related stuff will be moved to another PR

TuomasBorman (approved):
Great, thank you!

PR #840, opened 2026-01-27
refactor: switch style workflow to use git-auto-commit

artur-sannikov:
Use [this action](https://github.com/stefanzweifel/git-auto-commit-action) now
instead of custom script.

Also upgraded checkout Action to v6, and did some yaml styling.


TuomasBorman (approved):
Seems good, thanks!

Issue #803, opened 2025-09-11
mikropml update


TuomasBorman:
mikropml package will be soon updated: https://github.com/SchlossLab/mikropml/pull/349
Once the PR is merged, the package will have TreeSE support.

Update the ML chapter so that it utilizes the TreeSE functionality. These modifications could be done already; just use the fork.

After updating the OMA chapter, notify mikropml people so that they can add OMA link to their documentation.

@Benjamin-Valderrama Would you be able to implement this?

Benjamin-Valderrama:
Hi @TuomasBorman, I'm sorry for the late reply.

Thanks for letting me know. I'm a bit busy RN but happy to come back to this as soon as find some free time! 

Sorry for any inconvenience.

TuomasBorman:
@Benjamin-Valderrama Updates on this?

Benjamin-Valderrama:
Hi @TuomasBorman . I'm sorry, I completely forgot about this. I'll submit the PR this weekend.

TuomasBorman:
@Benjamin-Valderrama Would you be able to check the error in xgboost, and how it can be fixed in OMA? There was different error earlier related to xgboost. I think the current one is also related to xgboost, but unfortunately I do not have time to check it in detail.

I updated my local xgboost and was able to reproduce the error (with older version, I did not experience error):

```
> #| label: 4.3 - XGBoost
> 
> # Train XGBoost for classification
> xgb <- run_ml(
+     dataset = tse,
+     method = "xgbTree",
+     outcome_colname = "disease",
+     altexp = "prep_classification",
+     assay.type = "prep_classification",
+     col.var = alpha_divs_classification, # use these variables as predictors too
+     seed = 1,
+     kfold = 2,
+     cv_times = 2,
+     training_frac = .8,
+     find_feature_importance = FALSE
+ )
Something is wrong; all the AUC metric values are missing:
    logLoss         AUC          prAUC        Accuracy       Kappa           F1       Sensitivity   Specificity  Pos_Pred_Value
 Min.   : NA   Min.   : NA   Min.   : NA   Min.   : NA   Min.   : NA   Min.   : NA   Min.   : NA   Min.   : NA   Min.   : NA   
 1st Qu.: NA   1st Qu.: NA   1st Qu.: NA   1st Qu.: NA   1st Qu.: NA   1st Qu.: NA   1st Qu.: NA   1st Qu.: NA   1st Qu.: NA   
[...]

PR #841, opened 2026-01-28
downgrade xgboost

Benjamin-Valderrama:
Downgrade xgboost to version 1.7.7.1 to solve #803 .

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,7 +22,6 @@ biocViews:
 Depends:
     R (>= 4.0.0)
 Imports:
-    xgboost (>= 1.7.1.7)
 Suggests:
     ALDEx2,
     ANCOMBC,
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -22,6 +22,7 @@ biocViews:
 Depends:
     R (>= 4.0.0)
 Imports:
+    xgboost (>= 1.7.1.7)
```
TuomasBorman: Does it work in Suggests? Does this version solve the problem (the current version is larger than this, I guess)

TuomasBorman: You can also force exact version so it probably is possible to also define <= 1.7.x.x

Benjamin-Valderrama: I'm sorry, I don't know why I made that mistake. I meant == instead of >=. Besides, the correct version is the one in the PR description, not the one I actually wrote in the file. I've just sent an update with the right versions and this time in 'Suggests'.

TuomasBorman (approved):
Thanks, seems good! Hopefully it works.

Issue #842, opened 2026-01-31
bump OSCA URL


0xMuluh:
OSCA moved from https://bioconductor.org/books/release/OSCA/ to https://bioconductor.org/books/3.22/OSCA.basic/

TuomasBorman:
The first is the landing page for all OSCA books. The second one is version-specific, I would keep the first one as then we have to update the version to newest one manually

TuomasBorman:
At least it seems so. Please, let me know if I'm wrong

PR #843, opened 2026-01-31
Update link to clustering information in community_similarity.qmd

0xMuluh:
ping #842


TuomasBorman:
[> The first is the landing page for all OSCA books. The second one is version-specific, I would keep the first one as then we have to update the version to newest one manually](https://github.com/microbiome/OMA/issues/842#issuecomment-3871458475)

What was the justification for changing this? I might have missed something...

antagomir:
@0xMuluh could you comment on the motivation, let's then decide on merge

0xMuluh:
image

The motivation was because the initial redirect screen may be confusing to users. @TuomasBorman is right on added work of manual updates if this is now changed.

TuomasBorman (changes requested):
How about using link: https://bioconductor.org/books/release/OSCA.basic/clustering.html

PR #846, opened 2026-03-03
Add mention of binning transformation

raivo-otus:
Add lit ref and mention of binning transformation available in mia to the relevant OMA chapter.


TuomasBorman (approved):
Thanks!

PR #854, opened 2026-04-21
Update biocbook workflows

artur-sannikov:
Some actions use nodejs 20, which is deprecated. I also updated the versions of other workflows.

That's usually the work of [DependaBot](https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/dependabot-quickstart-guide) or Renovate.


antagomir:
I am merging as this seems straightfwd and doesnt conflict with other (so it easy to pull back if need be)

PR #871, opened 2026-07-28
Simplify MOFA2 pipeline with mofa2 wrapper

RiboRings:
With this PR, the chapter on MOFA2 is updated to reflect the streamlined MOFA2 pipeline using the `mofa2` wrapper that was recently implemented with this PR: https://github.com/bioFAM/MOFA2/pull/144.

### DIFF
--- a/inst/pages/multiassay_ordination.qmd
+++ b/inst/pages/multiassay_ordination.qmd
@@ -35,8 +35,11 @@ We use the `r BiocStyle::Biocpkg("MOFA2")` package for analysis.
 #| label: mofa_data1
 
 library(mia)
-# Load the data
-data(HintikkaXOData, package = "mia")
+
+# Import dataset
+data("HintikkaXOData", package = "mia")
+
+# Store as mae
 mae <- HintikkaXOData
 ```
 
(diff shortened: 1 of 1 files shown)


RiboRings:
@TuomasBorman @antagomir This is ready to merge

inst/pages/multiassay_ordination.qmd
```diff
@@ -35,108 +35,111 @@ We use the `r BiocStyle::Biocpkg("MOFA2")` package for analysis.
...
-# For simplicity, classify all high-fat diets as high-fat, and all the low-fat
-# diets as low-fat diets
-colData(mae)$Diet <- ifelse(
-    colData(mae)$Diet == "High-fat" | colData(mae)$Diet == "High-fat + XOS",
+
+# For simplicity, reduce samples to two categories: high vs low fat diet
+mae$Diet <- ifelse(
+    mae$Diet %in% c("High-fat", "High-fat + XOS"),
     "High-fat", "Low-fat"
 )
```
TuomasBorman, line 63: This was already in the old code, but there is "Fat" column in metadata. Why not use that?

inst/pages/multiassay_ordination.qmd
```diff
@@ -35,108 +35,111 @@ We use the `r BiocStyle::Biocpkg("MOFA2")` package for analysis.
...
-    mae,
-    groups = "Diet",
-    extract_metadata = TRUE
-)
-model
 ```
 
-Model options can be defined as follows:
+Note that agglomeration by taxonomic rank might produce duplicate feature names,
+such as "uncultured". If present, they can be made unique as follows:
```
TuomasBorman, line 103: Should this be handled by default, so that there should  not be duplicated names?

TuomasBorman (changes requested):
Looks good! Couple points (see comments)

inst/pages/multiassay_ordination.qmd
RiboRings: Thanks, updated.

inst/pages/multiassay_ordination.qmd
RiboRings: Do you mean model mofa should return an error/warning if names are duplicated? Or making them unique inside the mofa function?

I think users would better off make the names unique before running mofa, otherwise the link between original names and unique names is lost and users cannot trace the results back to their original set of features.

TuomasBorman: Sorry, I checked this quickly.

I meant that

agglomerateByPrevalence() should output unique feature names by default

PR #872, opened 2026-07-29
fix(community_similarity): use official vegan docs

artur-sannikov:
https://www.rdocumentation.org belongs to Datacamp.


antagomir:
Can you give a bit more context on this PR?

artur-sannikov:
Currently, we are using https://www.rdocumentation.org/, which belongs to [Datacamp](https://www.datacamp.com/). I would use the official documentation of the `vegan` package instead of a website maintained by a third-party company.

TuomasBorman (approved):

PR #875, opened 2026-07-29
feat(ci): strip trailing whitespaces

artur-sannikov:
Similar to how we do it [here](https://github.com/microbiome/OMAManuscript/tree/devel).


artur-sannikov:
Just realized, I already had this PR here (https://github.com/microbiome/OMA/pull/864).

PR #880, opened 2026-09-01
Remove deprecated workflow telemetry action

### LINKED ISSUE #873: Node.js 20 is deprecated
We get a warning about node.js 20 being deprecated.

It comes from this action: https://github.com/catchpoint/workflow-telemetry-action/. There is an [open issue ](https://github.com/catchpoint/workflow-telemetry-action/issues/107)to bump it to node.js 24, but the repo seems to be abandoned.

OSTA book disabled [this action](https://github.com/lmweber/OSTA/blob/devel/.github/workflows/biocbook.yml#L67) for time being.

It does not break the builds, but it's something to keep in mind.

0xMuluh:
Closes #873

## Summary

- Remove the optional workflow telemetry action that still declares Node.js 20.
- Leave the remaining book-build workflow unchanged.

## Validation

- YAML parsed successfully with `yaml::read_yaml()`.
- `git diff --check` passed.
- Confirmed no telemetry references remain.
- `actionlint` was not available locally.

### DIFF
--- a/.github/workflows/biocbook.yml
+++ b/.github/workflows/biocbook.yml
@@ -18,9 +18,6 @@ jobs:
       - name: 🧾 Checkout repository
         uses: actions/checkout@v6
 
-      - name: ⏳ Collect Workflow Telemetry
-        uses: runforesight/workflow-telemetry-action@v2
-
       - name: 🧹 Clean up Docker environment
         run: |
           docker system prune --all --force --volumes


TuomasBorman (changes requested):
Could you comment it out instead? Otherwise, we do not probably remember to add it back.

PR #881, opened 2026-09-01
Use scaled fitted values for DMM drivers

### LINKED ISSUE #737: Didichlet multinomial mixture example
https://github.com/microbiome/OMA/blob/ae6e8b05e04d96820f0bcc71d16d806117214023/inst/pages/clustering.qmd#L265

The following example visualizes unscaled values. Should be changed to something like this:

```
# Get the estimates on how much each phyla contributes on each cluster
models <- metadata(altExp(tse, "prevalent"))$DMM
# Get the number of clusters in the best model
n_clusters <- models[["best"]]
# Get the best model
best_model <- models[["dmm"]][[n_clusters]]

# Get the drivers from the model
drivers <- fitted(best_model, scale = TRUE)

# Plot by utilizing miaViz::plotLoadings
plotLoadings(drivers, ncomponents = 2)
```

0xMuluh:
## Summary
- select the best DMM model from metadata
- use the public scaled fitted values for loading visualization

## Validation
- `git diff --check`
- full execution unavailable because of a pre-existing local readr ABI mismatch

Closes #737

### DIFF
--- a/inst/pages/clustering.qmd
+++ b/inst/pages/clustering.qmd
@@ -304,11 +304,13 @@ features.
 #| fig-height: 2
 
 # Get the estimates on how much each phylum contributes on each cluster
-best_model <- metadata(altExp(tse, "prevalent"))$DMM$dmm[2]
-drivers <- as.data.frame(best_model[[1]]@fit$Estimate)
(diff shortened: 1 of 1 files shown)


TuomasBorman (approved):
Thanks!

PR #882, opened 2026-09-01
Correct zero-variance feature filtering

### LINKED ISSUE #822: Confusion/Error on filtering out zero-variance features
See comments [here.](https://github.com/microbiome/OMA/discussions/652#discussioncomment-14872869)

Chapter 10.5 Filtering out zero-variance features has some issues. 

The logic presented is flawed, and essentially removes low-variance features. 
Additionally the use of raw counts for the calculation of SD should be reassessed; Is it sound?

The text and code also needs polishing as currently the almost the same paragraph is repeated twice in succession and there is a unused variable (`th <- 1`).

0xMuluh:
## Summary
- correct the zero-variance filter to use the untransformed SD threshold of zero
- clarify the histogram scale and counts-assay scope
- remove the unused threshold variable

## Validation
- `git diff --check`
- affected R chunks parse successfully

Closes #822

### DIFF
--- a/inst/pages/subsetting.qmd
+++ b/inst/pages/subsetting.qmd
@@ -245,7 +245,8 @@ comparisons but also helps our machine learning models learn from
 meaningful data without additional noise.
 
 To filter these features, we begin by calculating their standard
-deviation. Then we visualize the variances with histogram.
+deviation across samples. Then we visualize the standard deviations with a
+histogram.
 
 ```{r}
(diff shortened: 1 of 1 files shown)


TuomasBorman:
@0xMuluh Could you

1. Check what filtering functions we have in mia
2. Add variance-filtering functionality (either own function or it could be added to existing one)

This PR could then use the function

PR #883, opened 2026-09-01
Document addMDS rarefaction feature retention

### LINKED ISSUE #824: `addMDS` drops smallest sample when `sample = min(colSums(assay(tse, "counts")))`
Reported by Anna Kaisanlahti:  
- `addAlpha(sample = min(colSums(assay(tse, "counts")))` keeps the smallest-depth sample.  
- `addMDS(sample = min(colSums(assay(tse, "counts")))` drops that sample unless `sample` is set noticeably lower.

Example: smallest sample depth = 5887 reads; `addMDS(sample = 5887)` drops it, but `sample = 5390` keeps it. This suggests inconsistent subsampling logic between `addAlpha()` and `addMDS()` (or their underlying helpers).

## Steps to reproduce

```r
min_reads <- min(colSums(assay(tse, "counts")))

# 1) Alpha: smallest sample retained
tse_alpha <- addAlpha(tse, sample = min_reads, niter = 100)

# 2) Beta: check if smallest sample still present
tse_mds <- addMDS(tse, sample = min_reads, niter = 100)
```

## Expected behavior

`addMDS()` should treat `sample` consistently with `addAlpha()` / `rarefyAssay()`: using the true minimum depth should not drop that smallest sample.

## Actual behavior

`addMDS()` excludes the smallest sample when `sample = min_reads`; only lower values keep it.

## Request

- Confirm if this behavior is intentional.
- If not, align `addMDS()` with `addAlpha()` and/or add a clear warning when `sample` will cause sample exclusion.

0xMuluh:
## Summary
- retain all features when rarefying in the addMDS example
- explain the interaction between `ntop` and rarefaction depth

## Validation
- `git diff --check`

Closes #824

### DIFF
--- a/inst/pages/community_similarity.qmd
+++ b/inst/pages/community_similarity.qmd
@@ -505,6 +505,7 @@ tse <- addMDS(
     method = "euclidean",
     niter = 10, # Number of iterations
     sample = min(colSums(assay(tse, "counts"))), # Rarefaction depth
+    ntop = Inf, # Keep all features when determining rarefaction depth
     transf = clr, # Applied transformation
     replace = TRUE,
     name = "MDS_aitchison_rarefied"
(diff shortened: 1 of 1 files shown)


inst/pages/community_similarity.qmd
```diff
@@ -505,6 +505,7 @@ tse <- addMDS(
     method = "euclidean",
     niter = 10, # Number of iterations
     sample = min(colSums(assay(tse, "counts"))), # Rarefaction depth
+    ntop = Inf, # Keep all features when determining rarefaction depth
```
TuomasBorman, line 508: This explanation is not 100% true. ntop is paramter in scater::runMDS. It selects the most variable features, by default 500 (ntop=500).

TuomasBorman (changes requested):
I think the correct place is to specify `ntop = Inf` in mia::addMDS (or is it possible to specify `subset_row=FALSE`? That would be more explicit).

Btw the same problem is in runPCA. We could amke mia alias for that (getPCA and addPCA)

PR #885, opened 2026-09-01
Clarify IntegratedLearner sample requirements

### LINKED ISSUE #817: Clarification for IntegratedLearner preprocessing
Currently, it's unclear in the documentation that [IntegratedLearner](https://microbiome.github.io/OMA/docs/devel/pages/integrated_learner.html#import-and-preprocess-data) requires both overlapping samples and matching sample names across experiments.

**Request**

1. Clarify that the mae object must contain complete cases across experiments.
2. Explicitly state that sample names must match between experiments character-for-character.
3. Provide a method to harmonize sample names, e.g., using sampleMap or renaming with patient IDs.

This would prevent confusion and save time when preparing data for I`ntegratedLearner/FromMAE().`

ping: https://github.com/himelmallick/IntegratedLearner/issues/45

0xMuluh:
## Summary
- document complete-case and exact sample-name requirements
- clarify that sampleMap does not impute or harmonize IntegratedLearner inputs
- add a preflight sample-name validation example

## Validation
- `git diff --check`

Closes #817

### DIFF
--- a/inst/pages/extra_material/integrated_learner.qmd
+++ b/inst/pages/extra_material/integrated_learner.qmd
@@ -54,6 +54,30 @@ mae <- MultiAssayExperiment(
 mae
 ```
 
+`IntegratedLearnerFromMAE()` requires complete cases across the selected
+experiments. Every experiment must contain the same samples, and the sample
+names must match character-for-character. A `sampleMap` records relationships
+between samples and subjects in a `MultiAssayExperiment`, but it does not
+impute missing measurements or harmonize names for IntegratedLearner.
+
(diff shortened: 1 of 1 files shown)


TuomasBorman (commented):
Is this really a requirement?

If it is so, `IntegratedLearner` should instead take the samples that are shared in all experiments, i.e., do something like this: https://github.com/microbiome/multiomics/blob/91318041e34f849fcdf4a639d1ed594d9c8e2b4b/R/utils.R#L466

I think it shouldn't block users to run the method.

PR #887, opened 2026-09-01
Handle mikropml assay and altExp outputs

### LINKED ISSUE #860: Clarification for ML data preprocessing
*This issue is forwarded from an outside source*

Original issue description;

> The function call preprocess_data() introduced in OMA in 21.3.1 Preprocess for classification task behaves a bit weirdly.
> I have four tse-objects that are subsets from the same container. With identical code, in three out of four preprocess_data() outputs “prep_classification” to altExp of the tse. In one of the four it outputs “prep_classification” to an assay directly to tse used, not to altExp.
> I rerun the code from scratch and without changing anything, 2/4 behaved the same way inputting the preprocessing only to assay.
> This seems to happen randomly in other datasets too, aka running the same script produces different behavior with the same tse objects.
> 
> I can do run_ml() on both occasions by setting the altExp to null if the preprocessed data is not in altExp.
> I’m having a consistent trouble with run_ml() with tse objects, and while the problem persist on either case, the error is only displayed when I use the preprocessed data from altExp, but not from the assay.

I am currently trying to coordinate and get the original code and data from the anonymous source to replicate the behavior.  

This is a function from the `mikropml` package, and the behavior is intended. The function removes features with zero/low variance, and in the case features don't match after removal the preprocessed data is placed in the altExp -slot of the tse. 

**Actionable steps**

We could make an **upstream issue**, as in discussion, it would be more clear to have all preprocessed data put into an altExp -slot regardless of feature removal causing mismatch. 

In the meantime **adding clarification** to `mikropml::preprocess_data()` functions use into the relevant OMA chapter would help clear confusion.

0xMuluh:
## Summary
- handle preprocessing results stored as assays or alternative experiments
- retrieve removed-feature metadata from the correct location
- pass the detected preprocessing location to `run_ml()`

## Validation
- `git diff --check`
- affected ML chunks parse successfully
- synthetic assay and altExp storage cases pass
- full mikropml workflow unavailable locally because mikropml is not installed

Closes #860

### DIFF
--- a/inst/pages/machine_learning.qmd
+++ b/inst/pages/machine_learning.qmd
@@ -223,9 +223,18 @@ tse <- preprocess_data(
     group_neg_corr = FALSE
 )
 
-# Identify the discarded features (that won't be used as predictors)
-prep_classification <- altExp(tse, "prep_classification")
-removed_feats <- metadata(prep_classification)[["removed_feats"]]
+# The result is stored in an altExp if features were removed, and otherwise
+# in an assay of the original TreeSE.
+prep_classification_altexp <- if (
+    "prep_classification" %in% altExpNames(tse)
+) "prep_classification" else NULL
+
+# Identify discarded features (that won't be used as predictors)
+removed_feats <- if (is.null(prep_classification_altexp)) {
+    metadata(tse)[["removed_feats"]]
+} else {
+    metadata(altExp(tse, prep_classification_altexp))[["removed_feats"]]
+}
 
 # If any alpha diversity index was removed, it will be discarded
 # from the predictors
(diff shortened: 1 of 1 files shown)


inst/pages/machine_learning.qmd
```diff
@@ -252,17 +265,31 @@ tse <- preprocess_data(
...
+} else {
+    NULL
+}
+
+# Identify discarded features (that won't be used as predictors)
+removed_feats <- if (is.null(prep_regression_altexp)) {
+    metadata(tse)[["removed_feats"]]
+} else {
+    metadata(altExp(tse, prep_regression_altexp))[["removed_feats"]]
+}
```
TuomasBorman, line 283: I think we should avoid these kind of scripts. They are working, but it makes the workflow more complex than necessary, which uses reader's focusing capacity.

inst/pages/machine_learning.qmd
```diff
@@ -252,17 +265,31 @@ tse <- preprocess_data(
     group_neg_corr = FALSE
 )
 
-# Identify the discarded features (that won't be used as predictors)
-prep_regression <- altExp(tse, "prep_regression")
-removed_feats <- metadata(prep_regression)[["removed_feats"]]
```
TuomasBorman, line 257: This is old code, but isn't the alpha diversity indices also in `altExp(assay())` as they are "preprocessed"?

TuomasBorman (changes requested):
I think, careful explanation is enough to solve the issue. The code itself should be "as simple as possible".

PR #888, opened 2026-09-01
Add HuMMANet resource reference

### LINKED ISSUE #879: Add HuMMANet
Add reference to databases and microbiome tools (supplement): https://www.biorxiv.org/content/10.64898/2026.08.24.746727v1

0xMuluh:
## Summary
- add a verified BibTeX citation for the HuMMANet preprint
- list HuMMANet under microbiome databases and tools resources

## Validation
- `git diff --check`
- verified title, author order, DOI, year, and URL against Crossref and bioRxiv

Closes #879

### DIFF
--- a/inst/assets/bibliography.bib
+++ b/inst/assets/bibliography.bib
@@ -3083,3 +3083,14 @@ doi = {10.1186/s13059-014-0550-8}
   urldate = {2026-01-12},
   year = {2026}
 }
+
+@Article{Verma2026,
+  author =  {Shivangi Verma and Nalin Arora and Chandra Prakash Ajay and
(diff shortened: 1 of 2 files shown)


TuomasBorman (changes requested):
Thanks!

Is it possible to fetch data as TreeSE (I cannot remember)? If not, the following page, might be better: https://github.com/microbiome/OMA/blob/devel/inst/pages/tool_list.qmd

This lists "all" microbiome related stuff

antagomir:
Yess, I didnt check this specific issue but let us keep in mind that in general we could cite external material instead of replicating stuff that already exists somewhere well prepared.

